PR #8, opened 2024-03-17
Build and install

eagraf:
A super basic build, release and install process for habitat. Here are the basics:
- Releases are created in the Github UI. When a release is created, a Github action is triggered that builds Habitat for all supported operating systems and chip architectures. When done, each binary is tarred into it's own archive, and then uploaded as an artifact to the release page. In addition, the latest version of an installer script is added as an artifact to the release. You can see an example release here: https://github.com/eagraf/habitat-new/releases/tag/v0.0.1-testing-14
- To install the latest release, run:
 ```
curl -sL https://github.com/eagraf/habitat-new/releases/latest/download/install.sh 2>&1 | bash
```
Note that `curl | bash` is bad security practice, we are just doing this for convenience.

This PR also fixes some bugs that become apparent when using the default value of `HABITAT_PATH`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,8 @@ on:
     branches: [ master ]
   pull_request:
     branches: [ master ]
+  release:
+    types: [ created ]
 
 jobs:
 
@@ -26,7 +28,7 @@ jobs:
       uses: golangci/golangci-lint-action@v4.0.0
 
     - name: Build
-      run: go build -v ./...
+      run: make build
 
     - name: Docker Build
       run: make docker-build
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: .github/workflows/ci.yml [hunk 1]
```diff
@@ -5,6 +5,8 @@ on:
     branches: [ master ]
   pull_request:
     branches: [ master ]
+  release:
+    types: [ created ]
```
arushibandi, line 9: what does `created` mean?

arushibandi (approved):

.github/workflows/ci.yml
eagraf: This is saying run the workflow when a release is created

PR #10, opened 2024-04-14
Remove fx

arushibandi:
Habitat is not yet complicated enough to need dependency injection. Typically systems with 100s of dependencies is when you start needing it, and we're nowhere close to that. Additionally, it makes debugging more complicated since you can't do a simple trace to see where failures originate and for me personally it decreases code readability the way we need to split up amongst files. This PR fully deletes fx, attempting the most minimal refactor in doing so.

[NOTE: Not ready for review, opening to check CI]


internal/node/reverse_proxy/proxy_test.go
```diff
@@ -40,81 +44,83 @@ func TestProxy(t *testing.T) {
...
-	proxy := &ProxyServer{
-		Rules: make(RuleSet),
-	}
+	proxy := NewProxyServer(logging.NewLogger(), &config.NodeConfig{})
 	err = proxy.Rules.Add("backend1", &RedirectRule{
 		Matcher:         "/backend1",
 		ForwardLocation: redirectedServerURL,
 	})
-	assert.Nil(t, err)
+	require.Nil(t, err)
```
eagraf, line 52: Why is this changing assert to require?

eagraf: https://www.yellowduck.be/posts/assert-vs-require-in-testify#:~:text=Which%20one%20you%20choose%20depends%20on%20the%20behaviour%20you%20are%20looking%20for.&text=With%20assert%20%2C%20you%20will%20get%20a%20log%20entry%20for%20each%20assertation.&text=With%20require%20%2C%20as%20soon%20as,the%20test%20will%20be%20skipped.

Oh nice I'm a fan, thanks for doing this

cmd/node/main.go
```diff
@@ -14,95 +15,80 @@ import (
...
+	if err != nil {
+		log.Fatal().Err(err)
+	}
+
+	hdbPublisher := pubsub.NewSimplePublisher[hdb.StateUpdate]()
+	db, dbClose, err := hdbms.NewHabitatDB(log, nodeConfig)
+	if err != nil {
+		log.Fatal().Err(err)
+	}
+	defer dbClose()
```
eagraf, line 34: Can we create an interface for these closing functions? I think it'd be good to enforce some consistency across them

internal/node/controller/controller.go
```diff
@@ -257,22 +266,20 @@ func (c *BaseNodeController) GetUserByUsername(username string) (*node.User, err
 
 // TODO this is basically a placeholder until we actually have a way of generating
 // the certificate for the node.
-func generateInitState(nodeConfig *config.NodeConfig) ([]byte, error) {
+func generateInitState(nodeConfig *config.NodeConfig) []byte {
```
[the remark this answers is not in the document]
eagraf: *actually looks like it was just changed heavilly, and this PR accidentally reverted some of those changes

cmd/node/main.go
```diff
@@ -14,95 +15,80 @@ import (
...
-	).Run()
+	ctx := context.Background()
+	log := logging.NewLogger()
+
+	nodeConfig, err := config.NewNodeConfig()
+	if err != nil {
+		log.Fatal().Err(err)
+	}
+
+	hdbPublisher := pubsub.NewSimplePublisher[hdb.StateUpdate]()
```
eagraf, line 29: This duplicates the instance of publisher [here](https://github.com/eagraf/habitat-new/pull/10/files#diff-702cab026ccaa869423a60af9169856f80bd6aab3f3b335e1a16cd69fd0198ffR16). Having two of these causes a deadlock because the channel is listening to one but pushing updates to another

cmd/node/main.go
```diff
@@ -14,95 +15,80 @@ import (
...
+		controller.NewMigrationRoute(nodeCtrl),
+	}
+
+	router := api.NewRouter(routes, log, nodeCtrl, nodeConfig)
+	proxy := reverse_proxy.NewProxyServer(log, nodeConfig)
+	tlsConfig, err := nodeConfig.TLSConfig()
+	if err != nil {
+		log.Fatal().Err(err)
+	}
+	proxyClose, err := proxy.Start(ctx, constants.DefaultPortReverseProxy, tlsConfig)
```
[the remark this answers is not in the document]
arushibandi: I generally try to pass a top-level context down through all the services, to chain them for cancelling. Might not be super necessary right now but I think that's the convention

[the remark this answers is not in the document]
arushibandi: Oh hmm yeah I just re-read that and we should probably only pass it through if we're waiting on a cancel signal for it somewhere in the server. we're not doing that yet so i'll do what you said.

cmd/node/main.go
arushibandi: nice catch, passed this into NewHabitatDb
do we have tests for that deadlock?

cmd/node/main.go
arushibandi: hmm I think it would just be `type closeFn func()`, but we can if you want!

internal/node/controller/controller.go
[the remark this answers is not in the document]
arushibandi: ah oops, will fix

cmd/node/main.go
```diff
@@ -14,95 +15,80 @@ import (
...
+	if err != nil {
+		log.Fatal().Err(err)
+	}
+	defer proxyClose()
```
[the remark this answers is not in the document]
arushibandi: Hm I looked this up and it said no ....
I think what we want is:
trap the SIGINT -> graceful shutdown (cancel top-level context that all downstreams are listening on)
if any other signal happens -> forcefully close

I'm down to do this in a separate PR since right now all the servers just forcefully call `Close()` rather than the `Shutdown(ctx)` , don't wanna block the other open PRs
Wanna go back through and do a deeper quality pass

cmd/node/main.go
eagraf: I was playing around with running tests with the race checker on. Idk if that will find deadlocks, especially since this is all in initialization code that's kinda hard to test. Maybe we can create an initialization check that could catch something like this

PR #11, opened 2024-04-15
Support atproto personal data server Pt 1.

eagraf:
This PR prepares Habitat to support running an [ATProto Personal Data Server](https://github.com/bluesky-social/pds). However, this PR does not have anything specifically related to the PDS application. Instead, this PR contains a bunch of improvements needed to run the PDS server in a subsequent PR.

New additions:
- Each app installation has a `driver_config` field, which configures that app's installation based off of which app driver it uses.
- Installing applications also sets up reverse proxy rules that will be used when the app is started. This lets us forward traffic to the PDS's `xrpc` endpoint from the Habitat API proxy url at `localhost:3001`
- Reverse proxy now does TLS termination
- Create a helper for generating the initial state of a node by migrating an empty state from version `v0.0.1` to the latest version. This future proofs our initial state code (e.g. avoid having an empty map for a field that shouldn't exist in the current version, rather than `nil`).
- Some bug fixes

This has already been tested with the ATProto PDS, there is just no reference to it in this PR yet. If you want to try it with the PDS, hit `https://localhost:3000/users/0/apps` with this body (make sure to change any local paths):
```
{
    "app_installation": {
        "name": "pds",
        "version": "1",
        "driver": "docker",
        "driver_config": {
            "env": [
                "PDS_DATA_DIRECTORY=/pds",
                "PDS_BLOBSTORE_DISK_LOCATION=/pds/blocks",
                "PDS_PLC_ROTATION_KEY_K256_PRIVATE_KEY_HEX=5290bb1866a03fb23b09a6ffd64d21f6a4ebf624eaa301930eeb81740699239c",
                "PDS_JWT_SECRET=bd6df801372d7058e1ce472305d7fc2e",
                "PDS_ADMIN_PASSWORD=password"
            ],
            "mounts": [
                {
                    "Type": "bind",
                    "Source": "/Users/ethan/code/habitat/habitat-new/.habitat/pds",
                    "Target": "/pds"
                }
            ],
            "exposed_ports": [ "5000" ],
            "port_bindings": {
                "3000/tcp": [
                    {
                        "hostIp": "127.0.0.1",
                        "hostPort": "5000"
                    }
                ]
            }
        },
        "registry_url_base": "registry.hub.docker.com",
        "registry_app_id": "ethangraf/pds",
        "registry_tag": "latest"
    },
    "reverse_proxy_rules": [
        {
            "type": "redirect",
            "matcher": "/xrpc",
            "target": "http://host.docker.internal:5000/xrpc"
        }
    ]
}
```

### DIFF
--- a/core/state/node/transitions.go
+++ b/core/state/node/transitions.go
@@ -3,6 +3,7 @@ package node
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/google/uuid"
@@ -39,6 +40,11 @@ func (t *InitalizationTransition) Patch(oldState []byte) ([]byte, error) {
 
 	if t.InitState.Processes == nil {
 		t.InitState.Processes = make(map[string]*ProcessState)
+	}
+
+	if t.InitState.ReverseProxyRules == nil {
+		rules := make(map[string]*ReverseProxyRule)
+		t.InitState.ReverseProxyRules = &rules
 	}
 
 	marshaled, err := json.Marshal(t.InitState)
@@ -154,6 +160,7 @@ func (t *AddUserTransition) Validate(oldState []byte) error {
 type StartInstallationTransition struct {
 	UserID string `json:"user_id"`
 	*AppInstallation
+	NewProxyRules []*ReverseProxyRule `json:"new_proxy_rules"`
 }
 
 func (t *StartInstallationTransition) Type() string {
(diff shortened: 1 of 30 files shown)


core/state/node/transitions.go
```diff
@@ -41,6 +42,11 @@ func (t *InitalizationTransition) Patch(oldState []byte) ([]byte, error) {
 		t.InitState.Processes = make(map[string]*ProcessState)
 	}
 
+	//	if t.InitState.ReverseProxyRules == nil {
```
arushibandi, line 45: remove commented

arushibandi (approved):
took a brief look only but don't want to block
can we add some comments about driver config?

PR #13, opened 2024-04-30
Frontend setup

sashankg:
Setup [Next.js](https://nextjs.org/) project with [static exports](https://nextjs.org/docs/app/building-your-application/deploying/static-exports) to be served by node. Using [Tailwind](https://tailwindcss.com/) for styling.

### DIFF
--- a/frontend/.eslintrc.json
+++ b/frontend/.eslintrc.json
@@ -1,0 +1,3 @@
+{
+  "extends": "next/core-web-vitals"
+}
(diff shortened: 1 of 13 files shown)


sashankg:
Not hooking it up to the habitat node server yet since we need to use `PathPrefix` ([source](https://github.com/gorilla/mux?tab=readme-ov-file#static-files)) which the current `Route` interface doesn't support. I'll wait for https://github.com/eagraf/habitat-new/pull/10 to merge to update the interfaces. 

Also I'm in favor of switching to the standard lib mux: https://pkg.go.dev/net/http#ServeMux

arushibandi (approved):

PR #14, opened 2024-05-09
Switch to standard lib ServeMux

sashankg:
Removes unnecessary dependency + makes it easier to implement serving static directory for FE. 

Also refactored the handlers tests to directly call the handler instead of setting up a server to route the requests to better isolate.

### DIFF
--- a/internal/node/controller/routes.go
+++ b/internal/node/controller/routes.go
@@ -9,7 +9,6 @@ import (
 	"github.com/eagraf/habitat-new/core/state/node"
 	"github.com/eagraf/habitat-new/internal/node/constants"
 	"github.com/eagraf/habitat-new/internal/node/hdb"
-	"github.com/gorilla/mux"
 	"golang.org/x/mod/semver"
 )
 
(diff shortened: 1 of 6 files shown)


eagraf (commented):
Looks great. Should we wait for #10, or merge this first?

sashankg:
> Looks great. Should we wait for #10, or merge this first?
 
Yeah let's wait just to keep things easy

internal/node/controller/routes.go
```diff
@@ -234,11 +217,6 @@ func (h *AddUserRoute) Method() string {
 }
 
 func (h *AddUserRoute) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	if r.Method != http.MethodPost {
```
arushibandi, line 237: why did these checks get deleted?

arushibandi (commented):
I think using `gorilla` is pretty standard and has some advanced routing feature that we might want in the future, but down to merge this for now

internal/node/controller/routes.go
arushibandi: oh nvm see this got moved to `processRoute`

arushibandi (approved):

PR #15, opened 2024-05-16
Start pds automatically

eagraf:
* Refactors node and hdb startup a bit. Instead of passing an initial state, an initial list of transitions is passed in.
* Also pins a toolchain version because that was causing issues


sashankg:
Didn't see this before opening https://github.com/eagraf/habitat-new/pull/17 but you wanna land #17 separately and rebase.

internal/node/controller/init.go
```diff
@@ -0,0 +1,117 @@
...
+			Version: "1",
+			UserID:  constants.RootUserID,
+			Package: node.Package{
+				Driver: "docker",
+				DriverConfig: map[string]interface{}{
+					"env": []string{
+						"PDS_HOSTNAME=ethangraf.com",
+						"PDS_DATA_DIRECTORY=/pds",
+						"PDS_BLOBSTORE_DISK_LOCATION=/pds/blocks",
+						"PDS_PLC_ROTATION_KEY_K256_PRIVATE_KEY_HEX=5290bb1866a03fb23b09a6ffd64d21f6a4ebf624eaa301930eeb81740699239c",
```
arushibandi, line 33: hmm should this key be in here?

arushibandi (commented):
one question

internal/node/controller/init.go
eagraf: Yeah I don't want to spend too much time about getting the configuration of PDS perfect right now.

arushibandi (approved):

PR #16, opened 2024-05-19
Integrate tsnet

eagraf:
This PR integrates the [Tailscale tsnet](https://tailscale.com/kb/1244/tsnet#serverlogger) library with Habitat, removing the need for a separate docker container to connect to the Tailnet.

Use of `tsnet` is optional. If the user doesn't pass in `TS_AUTHKEY` to the Habitat node, the node will start the reverse proxy using a regular TCP listener. If `TS_AUTHKEY` is specified, this registers the node as a machine on the corresponding Tailnet, and registers the host `habitat` on that network. For example, the node state endpoint under the reverse proxy is available at that host using a URL looking like this: `https://habitat:3001/habitat/api/node`

This PR also does a bunch of other stuff to support this change:
* Upgrade to Go 1.22. This was needed because the latest versions of `tsnet` relying on Go 1.21 had build issues due to relying [on the wrong version of Gvisor.](https://github.com/google/gvisor/commit/5b4fbd06d814e55c3b7cca8dec0a449454e464ea) Since we're behind anyways, might as well upgrade to Go 1.22 and fix the issue with `tsnet`
* Disabled authentication through mTLS for now. This workflow is too annoying for now, and adds very little value until we work out what the actual auth system will look like. Configuring certificates for multiple hosts on the client side would have been really annoying.
* Add debug logging level to suppress really noisy `tsnet` logs. By default, info level is used.
* Additional environment can be added to a gitignored `dev.env` file. I used this to set my `TS_AUTHKEY`. This is easier than supplying it before every command


sashankg:
Didn't see this before opening https://github.com/eagraf/habitat-new/pull/17 but you wanna land #17 separately and rebase.

eagraf:
> Didn't see this before opening #17 but you wanna land #17 separately and rebase.

Sounds good

.gitignore
```diff
@@ -9,3 +9,6 @@ coverage.out
 
 # Bin directory
 bin/
+
+# Developer specific environment
+dev.env
```
arushibandi, line 14: How do we generate this?

Makefile
```diff
@@ -43,8 +44,12 @@ install:: $(DEV_HABITAT_PATH)/habitat.yml $(CERT_DIR)/dev_node_cert.pem $(CERT_D
 docker-build:
 	docker build -t habitat_node -f ./build/node.dev.Dockerfile .
 
-run-dev:
-	TOPDIR=$(TOPDIR) DOCKER_WORKDIR=$(DOCKER_WORKDIR) DEV_HABITAT_PATH=$(DEV_HABITAT_PATH) TS_AUTHKEY=$(origin TS_AUTHKEY) docker-compose -f ./build/compose.yml up
+run-dev: $(PDS_BLOB_DIR)
```
arushibandi, line 47: is this supposed to be `PDS_BLOB_DIR=$(PDS_BLOB_DIR)`?

build/node.dev.Dockerfile
```diff
@@ -28,6 +28,7 @@ RUN mkdir -p /go/src/github.com/eagraf/habitat-new/pkg
 EXPOSE 3000
 EXPOSE 3001
 EXPOSE 4000
+EXPOSE 80
```
arushibandi, line 31: is this for `tailscale`? I wonder if we can specify different ports

cmd/node/main.go
```diff
@@ -44,24 +48,29 @@ func main() {
...
-	router := api.NewRouter(routes, log, nodeCtrl, nodeConfig)
-	proxy := reverse_proxy.NewProxyServer(log, nodeConfig)
+	router := api.NewRouter(routes, logger, nodeCtrl, nodeConfig)
+	proxy := reverse_proxy.NewProxyServer(logger, nodeConfig)
 	tlsConfig, err := nodeConfig.TLSConfig()
 	if err != nil {
 		log.Fatal().Err(err)
 	}
-	proxyClose, err := proxy.Start(constants.DefaultPortReverseProxy, tlsConfig)
+	proxyClose, err := proxy.Start(":"+constants.DefaultPortReverseProxy, tlsConfig)
```
arushibandi, line 57: ah yeah I fixed this in my PR too :/

internal/node/config/config.go
```diff
@@ -169,6 +200,11 @@ func (n *NodeConfig) TailscaleAuthkey() string {
 	return viper.GetString("tailscale_authkey")
 }
 
+func (n *NodeConfig) TailScaleStatePath() string {
+	// Note: this is intentionally not configurable for simplicity's sake.
+	return filepath.Join(n.HabitatPath(), "tailscale_state")
```
arushibandi, line 205: hm this defaults to `ts_state`

internal/node/controller/init.go
```diff
@@ -27,7 +27,7 @@ func generatePDSAppConfig(nodeConfig *config.NodeConfig) types.PostAppRequest {
 				Driver: "docker",
 				DriverConfig: map[string]interface{}{
 					"env": []string{
-						"PDS_HOSTNAME=ethangraf.com",
+						"PDS_HOSTNAME=localhost",
```
arushibandi, line 30: what does this do if it's set to `localhost` but uses your private key below?

internal/node/reverse_proxy/server.go
```diff
@@ -66,13 +70,40 @@ func (s *ProxyServer) Start(addr string, tlsConfig *tls.Config) (func(), error)
...
+			if err != nil {
+				return err
+			}
+			defer ln.Close()
+			listener = ln
+		}
+
 		if tlsConfig != nil {
-			err = httpServer.ListenAndServeTLS(s.nodeConfig.NodeCertPath(), s.nodeConfig.NodeKeyPath())
+			return httpServer.ServeTLS(listener, s.nodeConfig.NodeCertPath(), s.nodeConfig.NodeKeyPath())
```
arushibandi, line 103: I don't quite understand the differences between using `Serve` vs `ListenAndServe`?

eagraf: Pretty much no difference except that you pass the listener in directly rather than letting the function create it. We need it in this case because we build a custom listener with `tsnet`

internal/node/config/config.go
eagraf: Ahh I missed that there was another config for this

build/node.dev.Dockerfile
eagraf: I think this is leftover from when I was testing

Makefile
eagraf: No this is intentional. It just forces `make` to create that directory. I chose that one in particular because it builds all of the parent directories as well.

.gitignore
eagraf: Should have documented this, I made it manually. All I have in mine is `TS_AUTHKEY`. Just didn't want to keep pasting it in the shell

internal/node/controller/init.go
eagraf: The key below is to help with rotating certs, so has no effect on our current usage of the PDS. We will need to work out how to manage this properly in the long run, but not really high priority in my opinion right now.

arushibandi (approved):

PR #17, opened 2024-05-22
Upgrade Go v1.22

sashankg:
Prompted by https://github.com/eagraf/habitat-new/pull/14 where we need `http.Request.PathValue` which was added in 1.22

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/eagraf/habitat-new
 
-go 1.21
+go 1.22
 
 require (
 	github.com/docker/docker v25.0.1+incompatible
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/eagraf/habitat-new
 
-go 1.21
```
eagraf, line 3: I think we gotta run `go mod tidy` to recalculate the lock file

eagraf (approved):
I think we also need to get the coverage up because there is a bug in the coverage tool when using 1.21 that causes it to ignore directories that don't have tests. I added some ignores here: https://github.com/eagraf/habitat-new/pull/16/commits/52a475d5e8858049cb3eae3f6a92f0077bd9105c

go.mod
sashankg: Running doesn't seem to change anything

PR #18, opened 2024-05-26
context cleanup

arushibandi:
Some basic cleanup around contexts and http servers. We now trap `SIGINT` / `SIGTERM` and gracefully shutdown the reverse proxy / API server. Tested by sending `SIGINT` to the docker process running habitat. Additionally some simplification / deleting redundant code.


cmd/node/main.go
```diff
@@ -85,12 +73,132 @@ func main() {
...
+
+// serveFn takes in an http.Server and additional config and returns a callback that can be run in a separate go-routine.
+func serveFn(srv *http.Server, name string, opts ...option) func() error {
+	options := &serverOption{}
+	for _, o := range opts {
+		o(options)
+	}
+	return func() error {
+		var err error
+		if srv.TLSConfig != nil && options.tlsConfig != nil {
```
eagraf, line 191: Let's modify this to refer to `USE_TSNET` once this is merged: https://github.com/eagraf/habitat-new/pull/16/commits/9cae3e6efc4cdda0e6b96cc52b099773260620d7

cmd/node/main.go
```diff
@@ -85,12 +73,132 @@ func main() {
...
+	// Wait for the go-routines to finish
+	err = eg.Wait()
+	if err != nil {
+		log.Err(err).Msg("received error on eg.Wait()")
+	}
+	log.Info().Msg("Finished!")
+}
+
+// config provided to http.Server.ListenAndServeTLS()
+type tlsConfig struct {
```
eagraf, line 160: Why can't we just use the same type used by the `tls.Config` type?

eagraf (approved):

cmd/node/main.go
arushibandi: These are the cert & key file names, which is different from what's in `tls.Config`/

PR #22, opened 2024-06-18
Deployment fixes and embed frontend

eagraf:
This PR fixes some bugs with installation and makes the frontend embeddable inside the habitat node binary. The frontend is served with minimal additional setup when deployed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,6 +115,6 @@ $(TOPDIR)/bin/amd64-darwin/habitat-amd64-darwin.tar.gz: $(TOPDIR)/bin/amd64-darw
 
 # Embed the frontend in the binary
 internal/frontend/build:
-	cd $(TOPDIR)/frontend && npm run build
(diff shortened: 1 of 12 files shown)


Makefile
```diff
@@ -93,15 +98,22 @@ $(TOPDIR)/bin: $(TOPDIR)
 	mkdir -p $(TOPDIR)/bin
 
 # Linux AMD64 Builds
-$(TOPDIR)/bin/amd64-linux/habitat: $(TOPDIR)/bin
+$(TOPDIR)/bin/amd64-linux/habitat: $(TOPDIR)/bin internal/frontend/build
```
arushibandi, line 101: do we not still need `$(TOPDIR)/bin?

arushibandi (approved):

Makefile
eagraf: Some pre-existing rules still rely on it

PR #23, opened 2024-06-26
Eagraf/start pds after install

eagraf:
This PR has two refactors to state management, and then a change for starting the PDS automatically after it has been installed. There are also some changes to get development working on Linux.

1. Previously, `StateUpdate` only stored serialized JSON for the new state, as well as the state transition, which led to repeated bits of deserialization code sprinkled across all of our state update subscribers. This was refactored to have deserialization occur before the state update is sent to the subscribers, using the `StateUpdateInternal` interface.

2. Add an `Enrich` method to `Transition`. This allows for the system to populate data into the proposed state transition that was not specified by the client. This is helpful for generating timestamps and UUIDs, as well as denormalizing some contextual data into the transition data structure for easy access in downstream businesss logic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,8 @@ To run all the code and tooling in this repository, you will need to have the fo
 * golangci-lint: https://golangci-lint.run/usage/install/
 * go-test-coverage: Run `go install github.com/vladopajic/go-test-coverage/v2@latest`
 * Postman: https://www.postman.com/downloads/
+* pnpm: https://pnpm.io/installation
+* NodeJS: https://nodejs.org/en/download/package-manager
 
 ## Local Development
 The local dev setup runs the main Habitat node in a docker container. The code is built inside the container using [Air](https://github.com/cosmtrek/air), which allows for live reloading. To build the dev container, run:
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ make test-coverage
 ```
 This repository uses [gomock](https://github.com/uber-go/mock) to create mocks in tests. Generally, mocks will be generated with a command looking like this that writes the mock code into a `mocks` package sitting next to where the real code lives:
 ```
+// If you haven't installed before, run $go install go.uber.org/mock/mockgen@latest
 mockgen -source=internal/node/hdb/dbms.go -package mocks > internal/node/hdb/mocks/mock_dbms.go
 ```
 The mocks can be regenerated as needed when the interface they are mocking is changed. 
--- a/build/compose.yml
+++ b/build/compose.yml
@@ -33,4 +33,5 @@ services:
     env_file: ../dev.env
     depends_on:
       - habitat_frontend
-
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
--- a/build/compose.yml
+++ b/build/compose.yml
@@ -1,4 +1,4 @@
-name: habitat-node
+version: "3.3"
 services:
   habitat_frontend:
     image: habitat_frontend
@@ -30,9 +30,7 @@ services:
     build: 
       context: ..
       dockerfile: build/node.dev.Dockerfile
-    env_file:
-      - path: ../dev.env
-        required: false
+    env_file: ../dev.env
     depends_on:
       - habitat_frontend
 
--- a/core/state/node/schema.go
+++ b/core/state/node/schema.go
@@ -63,12 +63,20 @@ func (s NodeState) Bytes() ([]byte, error) {
 }
 
 func (s NodeState) GetAppByID(appID string) (*AppInstallationState, error) {
+	app, ok := s.AppInstallations[appID]
+	if !ok {
+		return nil, fmt.Errorf("app with ID %s not found", appID)
+	}
+	return app, nil
+}
+
+func (s NodeState) GetAppByName(appName string) (*AppInstallationState, error) {
 	for _, app := range s.AppInstallations {
-		if app.ID == appID {
+		if app.Name == appName {
 			return app, nil
 		}
 	}
-	return nil, fmt.Errorf("app with ID %s not found", appID)
+	return nil, fmt.Errorf("app with name %s not found", appName)
 }
 
 func (s NodeState) GetAppsForUser(userID string) ([]*AppInstallationState, error) {
--- a/core/state/node/schema.go
+++ b/core/state/node/schema.go
@@ -70,15 +70,6 @@ func (s NodeState) GetAppByID(appID string) (*AppInstallationState, error) {
 	return app, nil
 }
 
-func (s NodeState) GetAppByName(appName string) (*AppInstallationState, error) {
-	for _, app := range s.AppInstallations {
-		if app.Name == appName {
-			return app, nil
-		}
-	}
-	return nil, fmt.Errorf("app with name %s not found", appName)
-}
-
 func (s NodeState) GetAppsForUser(userID string) ([]*AppInstallationState, error) {
 	apps := make([]*AppInstallationState, 0)
 	for _, app := range s.AppInstallations {
--- a/core/state/node/schema.go
+++ b/core/state/node/schema.go
@@ -20,7 +20,7 @@ var nodeSchemaRaw string
 
 // TODO structs defined here can embed the immutable structs, but also include mutable fields.
 
-type NodeState struct {
+type State struct {
 	NodeID            string                           `json:"node_id"`
 	Name              string                           `json:"name"`
 	Certificate       string                           `json:"certificate"` // TODO turn this into b64
@@ -53,16 +53,16 @@ type ProcessState struct {
 	ExtDriverID string `json:"ext_driver_id"`
 }
 
-func (s NodeState) Schema() hdb.Schema {
+func (s State) Schema() hdb.Schema {
 	ns := &NodeSchema{}
 	return ns
 }
 
-func (s NodeState) Bytes() ([]byte, error) {
+func (s State) Bytes() ([]byte, error) {
 	return json.Marshal(s)
 }
 
-func (s NodeState) GetAppByID(appID string) (*AppInstallationState, error) {
+func (s State) GetAppByID(appID string) (*AppInstallationState, error) {
 	app, ok := s.AppInstallations[appID]
 	if !ok {
 		return nil, fmt.Errorf("app with ID %s not found", appID)
@@ -70,7 +70,7 @@ func (s NodeState) GetAppByID(appID string) (*AppInstallationState, error) {
 	return app, nil
 }
 
-func (s NodeState) GetAppsForUser(userID string) ([]*AppInstallationState, error) {
+func (s State) GetAppsForUser(userID string) ([]*AppInstallationState, error) {
 	apps := make([]*AppInstallationState, 0)
 	for _, app := range s.AppInstallations {
 		if app.UserID == userID {
@@ -80,7 +80,7 @@ func (s NodeState) GetAppsForUser(userID string) ([]*AppInstallationState, error
 	return apps, nil
 }
 
-func (s NodeState) GetProcessesForUser(userID string) ([]*ProcessState, error) {
+func (s State) GetProcessesForUser(userID string) ([]*ProcessState, error) {
 	procs := make([]*ProcessState, 0)
 	for _, proc := range s.Processes {
 		if proc.UserID == userID {
@@ -90,7 +90,7 @@ func (s NodeState) GetProcessesForUser(userID string) ([]*ProcessState, error) {
 	return procs, nil
 }
 
-func (s NodeState) GetReverseProxyRulesForProcess(processID string) ([]*ReverseProxyRule, error) {
+func (s State) GetReverseProxyRulesForProcess(processID string) ([]*ReverseProxyRule, error) {
 	process, ok := s.Processes[processID]
 	if !ok {
 		return nil, fmt.Errorf("process with ID %s not found", processID)
@@ -108,12 +108,12 @@ func (s NodeState) GetReverseProxyRulesForProcess(processID string) ([]*ReverseP
 	return rules, nil
 }
 
-func (s NodeState) Copy() (*NodeState, error) {
+func (s State) Copy() (*State, error) {
 	marshaled, err := s.Bytes()
 	if err != nil {
 		return nil, err
 	}
-	var copy NodeState
+	var copy State
 	err = json.Unmarshal(marshaled, &copy)
 	if err != nil {
 		return nil, err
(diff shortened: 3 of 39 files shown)


README.md
```diff
@@ -61,6 +61,7 @@ make test-coverage
 ```
 This repository uses [gomock](https://github.com/uber-go/mock) to create mocks in tests. Generally, mocks will be generated with a command looking like this that writes the mock code into a `mocks` package sitting next to where the real code lives:
 ```
+// If you haven't installed before, run $go install go.uber.org/mock/mockgen@latest
```
arushibandi, line 64: we should add this type of stuff to a script for running dev

build/compose.yml
```diff
@@ -1,4 +1,4 @@
-name: habitat-node
+version: "3.3"
```
arushibandi, line 1: where did name go?

build/compose.yml
```diff
@@ -30,9 +30,8 @@ services:
     build: 
       context: ..
       dockerfile: build/node.dev.Dockerfile
-    env_file:
-      - path: ../dev.env
-        required: false
+    env_file: ../dev.env
```
arushibandi, line 33: maybe we should rename this to not `dev` if it's in the actual release

build/compose.yml
```diff
@@ -30,9 +30,8 @@ services:
...
     depends_on:
       - habitat_frontend
-
+    extra_hosts:
```
arushibandi, line 36: ooh what is this for? can you leave a comment

core/state/node/schema.go
```diff
@@ -63,12 +63,20 @@ func (s NodeState) Bytes() ([]byte, error) {
...
+	if !ok {
+		return nil, fmt.Errorf("app with ID %s not found", appID)
+	}
+	return app, nil
+}
+
+func (s NodeState) GetAppByName(appName string) (*AppInstallationState, error) {
 	for _, app := range s.AppInstallations {
-		if app.ID == appID {
+		if app.Name == appName {
```
arushibandi, line 75: a later question, but what if there are multiple? we probably don't need to use this function except for showing UI stuff right?

core/state/node/transitions.go
```diff
@@ -53,6 +54,10 @@ func (t *InitalizationTransition) Patch(oldState []byte) ([]byte, error) {
 	}]`, marshaled)), nil
 }
 
+func (t *InitalizationTransition) Enrich(oldState []byte) error {
+	return nil
```
arushibandi, line 58: should this instead return error if it's unexpected for this to ever be called on init?

core/state/node/transitions.go
```diff
@@ -83,6 +88,10 @@ func (t *MigrationTransition) Patch(oldState []byte) ([]byte, error) {
 	return json.Marshal(patch)
 }
 
+func (t *MigrationTransition) Enrich(oldState []byte) error {
+	return nil
```
arushibandi, line 92: ^same thing for this
or should this include something?

core/state/node/transitions.go
```diff
@@ -109,25 +118,46 @@ func (t *MigrationTransition) Validate(oldState []byte) error {
...
-	}]`, t.UserID, t.UserID, t.Username, t.Certificate)), nil
+		"value": %s
+	}]`, t.EnrichedData.User.ID, user)), nil
+}
+
+func (t *AddUserTransition) Enrich(oldState []byte) error {
+
+	id := uuid.New().String()
+
+	t.EnrichedData = &AddUserTranstitionEnrichedData{
```
arushibandi, line 153: I'm a little confused by `Enrich` because it sounds like "metadata" or extra info ... why can't this data be generated at the time the transition is committed and then returned to the caller?  like maybe in an `Apply` method

core/state/node/transitions.go
```diff
@@ -109,25 +118,46 @@ func (t *MigrationTransition) Validate(oldState []byte) error {
...
+
+type AddUserTranstitionEnrichedData struct {
+	User *User `json:"user"`
 }
 
 func (t *AddUserTransition) Type() string {
 	return TransitionAddUser
 }
 
 func (t *AddUserTransition) Patch(oldState []byte) ([]byte, error) {
```
arushibandi, line 135: can we make these function signatures / states into separate types to be more clear about what the return value means?

internal/node/hdb/executor.go
```diff
@@ -10,13 +10,18 @@ import (
...
-	Restore        bool
+	Index      uint64
+	SchemaType string
+	DatabaseID string
+	Restore    bool
+
+	StateUpdateInternal
+}
+
+type StateUpdateInternal interface {
```
arushibandi, line 21: this confuses me because usually "internal" or "inner" refers to a struct definition while the outer one is the interface, so i think the naming feels kinda backwards. also what was the motivation for this change?

internal/node/package_manager/subscriber.go
```diff
@@ -49,22 +49,58 @@ func (e *InstallAppExecutor) Execute(update *hdb.StateUpdate) error {
...
+	err = e.nodeController.FinishAppInstallation(t.UserID, appInstallation.ID, appInstallation.RegistryURLBase, appInstallation.RegistryPackageID, t.StartAfterInstallation)
 	if err != nil {
 		return err
 	}
 
 	return nil
 }
 
+// FinishInstallExecutor is a state update executor that finishes the installation of an application.
+// Currently it is just run for it's PostHook, which will try to start the app if StartAfterInstallation is set.
```
arushibandi, line 70: why can't we just make a PostHook on `InstallApp`?

build/compose.yml
eagraf: This field is no longer supported in more recent versions of docker: https://stackoverflow.com/questions/73427249/name-does-not-match-any-of-the-regexes-x-while-trying-to-set-the-project

build/compose.yml
eagraf: This is just to get communication between containers working for Linux. On MacOS the docker internal host already pointed to the host machine by default, so it wasn't necessary before. I just got development in Linux working on this PR.

build/compose.yml
eagraf: I think this is fine for now, this compose file is only really used for development currently. Definitely worth a look later how we get critical configuration into the node when things are more mature.

core/state/node/schema.go
eagraf: This is leftover, but not currently used, I can remove it.

internal/node/package_manager/subscriber.go
eagraf: We don't want to run this unless the state for the app is already `installed` to maintain consistency. When the `PostHook` of `InstallAppExecutor` runs,[ it calls](https://github.com/eagraf/habitat-new/pull/23/files#diff-a5e50c6674dee457df8c8707dccd33e65ea6bfa9f2a38662b8247b927294111aR61) `nodeController.FinishAppInstallation`. With the way the controller works currently, it is making an aynchronous request to the state machine, so just because the function call completed successfully does not mean that the state was updated to `installed`.

We could create some sort of helper that makes state update requests to the state machine synchronous, but I think that would end up being more confusing than just creating another subscriber. The reason it would be more confusing is that Raft doesn't really provide a guarantee on how long the request will take to complete if there are multiple nodes involved, so we'd need to implement a timeout. I think timeouts should be avoided, because they can cause what appears to be inconsistent behavior to the client.

core/state/node/transitions.go
eagraf: This is just a no-op, I can comment that as well.

core/state/node/transitions.go
eagraf: Definitely open to suggestions on naming, I'm not sure if `Enrich` is the best we can do.

The main purpose it to clearly separate out the information needed to request a state update, from the schema for the data that actually makes up a given state update. Having these separated will help prevent a lot of bugs.

I need to think about how to return the data to users. It would be useful in some cases, but it runs into issues with the asynchronous way requests to Raft have to take place. If we make the function blocking so we can return the data after application, we'd could potentially wait indefinitely if Raft gets into a bad state. 

But, I think we can potentially return the data at the request time, which could be helpful in some situations.

internal/node/hdb/executor.go
eagraf: Main motivation was to stop having to  unmarshal state bytes in the consumers of the state update, which was getting annoying and making everything harder to read.

I'm gonna try to refactor this to be even simpler.

core/state/node/schema.go
```diff
@@ -63,12 +63,11 @@ func (s NodeState) Bytes() ([]byte, error) {
 }
 
 func (s NodeState) GetAppByID(appID string) (*AppInstallationState, error) {
```
arushibandi, line 65: nit: I think the convention is to name this just `(s State)` so that other packages can say `node.State` rather than `node.NodeState` which is redundant

internal/node/hdb/types.go
```diff
@@ -20,8 +20,18 @@ type TransitionWrapper struct {
...
+	// Type returns a string constant identifying the type of state transition. Currently these are in snake case.
 	Type() string
+
+	// Patch takes the old state and returns a JSON patch to apply to the old state to get the new state.
 	Patch(oldState []byte) ([]byte, error)
+
+	// Enrich adds important data to the transition that is not specified when submitted by the client.
+	// For example, the id of a newly created object can be generated during the enrichment phase. The id is not
+	// something we'd want the client to submit, but we want it to be included in the transition.
+	Enrich(oldState []byte) error
```
arushibandi, line 32: can we make `oldState []byte` -> `oldState State` and have `type State []byte`? that way we can change out the type underneath later if we need.

Also, why can't we move this itno `core/state/node/transitions`? it feels weird that this interface type is in `internal/node` when it's actually used in `core`

internal/node/package_manager/restore.go
```diff
@@ -11,17 +9,12 @@ type PackageManagerRestorer struct {
...
 
-func (r *PackageManagerRestorer) Restore(restoreEvent *hdb.StateUpdate) error {
-	var nodeState node.NodeState
-	err := json.Unmarshal(restoreEvent.NewState, &nodeState)
-	if err != nil {
-		return err
-	}
-
+func (r *PackageManagerRestorer) Restore(restoreEvent hdb.StateUpdate) error {
+	nodeState := restoreEvent.NewState().(*node.NodeState)
```
arushibandi, line 13: why is this getting cast? shouldn't the return type be `node.NodeState`?

internal/node/package_manager/subscriber.go
arushibandi: Hm okay it just feels weird that we're running something just for its PostHook, like it seems in that case should make something that doesn't have a `PostHook` and just run it when we want. Especially because this is called `FinishInstall`, it should do the thing it says.


for e.g. can we have an executor that always runs on node start up and and starts any apps that need to be started?

internal/node/pubsub/pubsub.go
```diff
@@ -10,14 +10,14 @@ type Event interface {
 }
 
 type Publisher[E Event] interface {
-	PublishEvent(*E) error
-	GetChan() <-chan *E
+	PublishEvent(E) error
```
arushibandi, line 13: what motivated this change?

eagraf: If the event is an interface, this implies passing a pointer to an interface, which is weird.

internal/node/package_manager/subscriber.go
eagraf: I think we can put this in the main execution for the subscriber if that makes it less confusing, there won't be any difference.

core/state/node/schema.go
eagraf: True, we can do this in a later PR since this one is pretty big already

internal/node/hdb/types.go
eagraf: I think a lot of the weirdess with types is that I'm trying to keep the state machine as generic as possible since we will eventually need it to handle different schemas (I'm thinking about community schemas down the line). But yeah a type alias here will help with readability. I'm thinking `SerializedState` because we already have `State` reserved.

internal/node/package_manager/restore.go
eagraf: For the same reason as above, there is weirdness because I want the state machine system to be reusable for community schemas down the line.

internal/node/controller/controller.go
```diff
@@ -23,10 +23,10 @@ type NodeController interface {
 	GetUserByUsername(username string) (*node.User, error)
 
 	InstallApp(userID string, newApp *node.AppInstallation, newProxyRules []*node.ReverseProxyRule) error
-	FinishAppInstallation(userID string, appID, registryURLBase, registryPackageID string) error
+	FinishAppInstallation(userID string, appID, registryURLBase, registryPackageID string, startAfterInstall bool) error
 	GetAppByID(appID string) (*node.AppInstallation, error)
 
-	StartProcess(process *node.Process) error
+	StartProcess(appID string) error
```
arushibandi, line 29: What's the difference between an app and process again? We have some internal processes that aren't apps but all apps are processes?

internal/node/controller/init.go
```diff
@@ -41,12 +41,12 @@ func generatePDSAppConfig(nodeConfig *config.NodeConfig) types.PostAppRequest {
 							Target: "/pds",
 						},
 					},
-					"exposed_ports": []string{"5000"},
+					"exposed_ports": []string{"5001"},
```
arushibandi, line 44: was this wrong before?

internal/node/hdb/types.go
```diff
@@ -20,12 +23,22 @@ type TransitionWrapper struct {
...
+	// For example, the id of a newly created object can be generated during the enrichment phase. The id is not
+	// something we'd want the client to submit, but we want it to be included in the transition.
+	Enrich(oldState SerializedState) error
+
+	// Validate checks that the transition is valid given the old state.
+	Validate(oldState SerializedState) error
 }
 
-func WrapTransition(t Transition, patch []byte, oldState []byte) (*TransitionWrapper, error) {
+func WrapTransition(t Transition, patch []byte, oldState SerializedState) (*TransitionWrapper, error) {
```
arushibandi, line 41: thanks for making this change!!

core/state/node/transitions.go
arushibandi: Hm yeah still not convinced having extra enriched data fields is the right move, but i understand the two different use cases now, i think it's fine for a while

arushibandi (approved):

internal/node/controller/controller.go
eagraf: App or `AppInstallation` is how we keep track of what's installed by the user. A process is just an instance of an app running. In the future, we may have processes that aren't tied to apps, but that doesn't exist right now

internal/node/controller/init.go
eagraf: This one is fun. On MacOS, AirPlay will occasionally listen on port 5000, making this break. So apparently, never use port 5000 https://forums.developer.apple.com/forums/thread/682332

PR #24, opened 2024-07-21
Eagraf/login

eagraf:
Basic login functionality:
* Adds a `login` and `add-user` page to the frontend. After a user has logged in, there 
* Logging in and registering are just wrappers around the PDS
* Some fixes to the reverse proxy as well
* Add support for Tailscale funnel and port 443


internal/node/controller/pds.go
```diff
@@ -0,0 +1,130 @@
...
+	"fmt"
+	"io"
+	"net/http"
+
+	types "github.com/eagraf/habitat-new/core/api"
+
+	"github.com/eagraf/habitat-new/internal/node/config"
+)
+
+type PDSClientI interface {
```
arushibandi, line 16: nit, but I think the naming here could be a little better. How do you feel about
```
type Client interface {} // remove "PDS" since other packages will already do 'pds.Client'

type ClientForNode struct {}
```

or something similar.
Also why does this need to be an interface?

internal/node/controller/routes.go
```diff
@@ -212,12 +212,66 @@ func (h *AddUserRoute) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
 
-	// TODO validate request
 	w.WriteHeader(http.StatusCreated)
+	w.Header().Set("Content-Type", "application/json")
+	w.Write(respBody)
+}
```
arushibandi, line 233: ^ should we be checking more `err`s here?

internal/node/reverse_proxy/proxy.go
```diff
@@ -64,6 +64,7 @@ type RuleHandler interface {
 	Type() string
 	Match(url *url.URL) bool
 	Handler() http.Handler
+	Rank() int
```
arushibandi, line 67: what do we need this for? I thought this could be implemented more natively, with the ordering in which we add rules?

eagraf: This was a quick hack. I think the way the reverse proxy is configured right now is very clunky, and I'm thinking about refactoring it. Can we defer until then?

internal/node/controller/pds.go
eagraf: The interface is for easy mocking. I have the `PDS` in front because this is still in the `controller` package. Ideally, we use the client from `indigo` once that is integrated into our repo

core/api/pds.go
```diff
@@ -0,0 +1,21 @@
...
+	Handle     string `json:"handle"`
+	Password   string `json:"password"`
+	InviteCode string `json:"inviteCode"`
+}
+
+type PDSCreateAccountResponse map[string]interface{}
+
+type PDSCreateSessionRequest struct {
+	Identifier string `json:"identifier"` // email or handle
+	Password   string `json:"password"`
```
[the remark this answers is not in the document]
eagraf: Definitely not the end state of this.  I don't want the scope of the PR to expand more right now. I think we can come back to security in the future.

internal/node/controller/routes.go
eagraf: What do you have in mind?

arushibandi: well, we should log an error if we're unable to write a response right?

arushibandi (approved):

PR #26, opened 2024-08-03
Eagraf/app frontends

eagraf:
This PR allows for app frontends to be served by the Habitat node. When hooked up to a reverse proxy file server, the frontend becomes available over a users tailscale network. 

Implementation:
* A new `web` driver is created for Habitat apps. Note that an app with separate backends and frontends will require two "habitat" apps to install. We can create wrapper transitions in the future that install the required Habitat apps for a single user-facing app.
* The `web` driver just downloads a `.tar.gz` of the bundled static contents of a website, and puts it under a versioned path in the `$HABITAT_PATH/web` directory. When a frontend is installed, a reverse proxy rule is also added.
* Just like `docker` based apps, `web` based apps need to be started and stopped. In this case, starting the app just turns on the reverse proxy rule, making it available to the user.

An example frontend app that is compatible with this system can be seen here: https://github.com/eagraf/blog-test
A frontend repository needs to implement the following to be compatible with this app installation method:
* It must be buildable into a static bundle, that can then be tarred and compressed. In this case, we use the `output: export` option in `next.config.mjs` to make sure it get's exported as a static bundle when built.
* The base path for the built static bundle needs to match the reverse proxy matching rule path we supply to habitat. This is because when the index file for the website is loaded, CSS, JS and other static assets from the bundle are loaded relative to the basepath. In the example, we use the `NextJS` `basePath` option in `next.config.mjs` to configure the basePath as `/blog-app`, resulting in a Habitat URL looking like `https://habitat:3001/blog-app`
* The frontend repo has a github action for building and releasing the static bundle. Essentially, it just runs `next build`, and then tars up the `out` directory, and uploads it to the release. An example release is here: https://github.com/eagraf/blog-test/releases/tag/test-release-10

Image of the frontend being hosted from a Habitat host:
Screen Shot 2024-08-03 at 4 00 52 PM


core/state/node/migrations/0000000005-web-driver.json
```diff
@@ -0,0 +1,26 @@
+{
+    "old_version": "v0.0.4",
+    "new_version": "v0.0.5",
+    "up": [
+        {
+            "op": "add",
+            "path": "/properties/app_installations/additionalProperties/properties/driver/enum/1",
```
arushibandi, line 7: btw, did you have to generate this yourself or do we have a less manual way of doing these?

internal/node/drivers/web/driver.go
```diff
@@ -0,0 +1,15 @@
+package web
+
+import (
+	"github.com/eagraf/habitat-new/internal/node/package_manager"
+)
+
+type WebBundleInstallationConfig struct {
```
arushibandi, line 7: same thing about naming on this one, remove 'Web'

internal/node/reverse_proxy/proxy.go
```diff
@@ -52,24 +58,27 @@ func (r RuleSet) AddRule(rule *node.ReverseProxyRule) error {
...
 	Handler() http.Handler
+	Rank() int
 }
 
 type FileServerRule struct {
 	Matcher string
 	Path    string
-	FS      fs.FS // Optional, instead of using Path, pass in an fs.FS. Useful for embedding the Habitat frontend.
+
+	FS       fs.FS  // Optional, instead of using Path, pass in an fs.FS. Useful for embedding the Habitat frontend.
```
arushibandi, line 80: should we use a `type Option` here? (like in https://pkg.go.dev/github.com/bradenaw/retry#Retry)

arushibandi (approved):

internal/node/constants/constants.go
```diff
@@ -12,6 +12,7 @@ const (
 
 	// App driver names
 	AppDriverDocker = "docker"
```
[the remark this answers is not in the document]
eagraf: I like having the common part be the prefix, so it's easier to search for.

core/state/node/migrations/0000000005-web-driver.json
eagraf: It's all manual right now. I have some code that generates the list of patches automatically, but it's only used for testing purposes. It should be easy to make it into a separate tool to generate these.

internal/node/reverse_proxy/proxy.go
eagraf: Good idea - I think that can go into the upcoming refactor for proxy stuff, since there are a couple other ways this is clunky.

PR #28, opened 2024-09-22
Eagraf/fishtail

eagraf:


Makefile
```diff
@@ -56,6 +56,7 @@ run-dev: $(PDS_BLOB_DIR)
 	DOCKER_WORKDIR=$(DOCKER_WORKDIR) \
 	DEV_HABITAT_PATH=$(DEV_HABITAT_PATH) \
 	DEV_HABITAT_APP_PATH=$(DEV_HABITAT_APP_PATH) \
+	HOST_MACHINE_REPO_PATH=$(TOPDIR) \
```
arushibandi, line 59: why do we need this? can't we just use `TOPDIR`?

internal/node/fishtail/publisher.go
```diff
@@ -0,0 +1,94 @@
...
+// for each subscriber that failed to receive the ingestion chain, and an overall error for unexpected
+// errors encountered while publishing the ingestion chain.
+func (sc *ATProtoEventPublisher) Publish(ingestionChain *IngestedRecordChain) ([]error, error) {
+	// When a certain record is modified, we want to publish it to the subscribers.
+	// In addition to that record, we traverse the chain of all the linked records
+	// and send them in the same message. That way, the various subscribers will
+	// have up-to-date info without the need to query the same data from thhe PDS.
+
+	// Find subscribers based on the collection of the initial record
+	subscribers := sc.GetSubscriptions(ingestionChain.Collection)
```
arushibandi, line 57: nit/style: one thing you could do is return `([]string, bool)` from `sc.GetSubscriptions where the `bool` indicates if there is something, or even an `int` indicating the length

then you could do 
```
subscribers, ok := ...
if !ok {
  return ...
}
```

just a little prettier to read

internal/node/fishtail/publisher.go
```diff
@@ -0,0 +1,94 @@
...
+		if resp.StatusCode != http.StatusOK {
+			body, _ := io.ReadAll(resp.Body)
+			errors = append(errors, fmt.Errorf("endpoint %s returned status %d: %s", target, resp.StatusCode, string(body)))
+			continue
+		}
+
+		log.Info().Msgf("Successfully sent %d records to ingest endpoint", len(ingestionChain.Records))
+	}
+
+	return errors, nil
```
arushibandi, line 93: what happens if there are errors? do we stop the fishtail? we should just log it right? (i didn't go back to see what happens in the call for this)

internal/node/fishtail/subscriber.go
```diff
@@ -0,0 +1,107 @@
...
+import (
+	"encoding/json"
+
+	"github.com/eagraf/habitat-new/core/state/node"
+	"github.com/eagraf/habitat-new/internal/node/config"
+	"github.com/eagraf/habitat-new/internal/node/controller"
+	"github.com/eagraf/habitat-new/internal/node/hdb"
+)
+
+func NewFishtailIngestSubscriber(
```
arushibandi, line 12: I want to be a bit careful with naming ... is this intended to be used by the fishtail service or by individual apps using fishtail? if the latter, then I don't think we should be pasing in `nodeConfig` and `nodeController` here. If the former, then isn't it subscribing to the firehose?

internal/node/fishtail/subscriber.go
```diff
@@ -0,0 +1,107 @@
...
+
+func (f *FishtailIngestExecutor) TransitionType() string {
+	return node.TransitionStartProcess
+}
+
+func (f *FishtailIngestExecutor) ShouldExecute(update hdb.StateUpdate) (bool, error) {
+	return true, nil
+}
+
+func (f *FishtailIngestExecutor) Execute(update hdb.StateUpdate) error {
```
arushibandi, line 47: I don't think this needs to be part of the whole state machine ... fishtail ingesting should be dynamic things that apps can susbcribe/unsubscribe to in the middle of the process, completely decoupled from node state.

internal/node/fishtail/ingest.go
```diff
@@ -0,0 +1,229 @@
...
+	if _, ok := ic.seen[uri.String()]; !ok {
+		ic.seen[uri.String()] = true
+		ic.toIngest <- uri
+		return true
+	}
+	return false
+}
+
+// IngestNext ingests the next record in the chain. If there are no more records to ingest, it returns false.
+func (ic *RecordChainIngester) IngestNext() (bool, error) {
```
arushibandi, line 112: I'm a little confused about what the "Chain" is here? Are there docs you can link to explaining why we need to do this?

arushibandi (commented):
Left a bunch of preliminary comments. Want to understand some of these a bit deeper then I'll go in and do another review more focused on the code

core/state/node/migrations.go
```diff
@@ -391,6 +391,35 @@ var NodeDataMigrations = MigrationsList{
...
+		down: func(state *State) (*State, error) {
+			newState, err := state.Copy()
+			if err != nil {
+				return nil, err
+			}
+
+			if newState.ReverseProxyRules != nil {
+				updatedRules := make(map[string]*ReverseProxyRule)
+				for id, rule := range *newState.ReverseProxyRules {
+					if rule.Type != ProxyRuleEmbeddedFrontend && rule.Type != ProxyRuleFishtailIngest {
```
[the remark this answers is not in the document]
eagraf: The JSON file specifies how the schema is migrated, but it doesn't actually say what to do about the data, which is why we need the go file. For example, imagine you split a `full_name` field to be two new fields called `first_name` and `last_name`. The JSON schema migration just says that the old field is deprecated, and add the two new fields, but it doesn't actually contain any logic for how legacy data would be converted to the new fields. That what the go migrations are for.

Overall, I'm not super happy that they are in go. I've written postgres migrations in `pgpsql` which let us do both in one place before, but it's a bit clunky. This area will definitely need some attention in the future.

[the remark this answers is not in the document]
eagraf: > having a hard time understanding the intention of this migration
it includes all rules from the old state except the embedded frontend and the fishtail ingest?
Yes

internal/node/appstore/routes_test.go
```diff
@@ -41,7 +41,8 @@ func TestRenderDevAppsList(t *testing.T) {
 
 	// Create a test node config
 	v := viper.New()
-	v.Set("habitat_path", "/home/fakeuser/.habitat")
+	v.Set("host_machine_repo_path", "/home/fakeuser/.habitat")
+	v.Set("environment", constants.EnvironmentDev)
```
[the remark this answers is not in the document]
eagraf: This test is specifically for making sure the app list is rendered a certain way when in the dev environment. The apps we see in `dev` will follow a different path than prod at the moment

internal/node/fishtail/README.md
```diff
@@ -0,0 +1,121 @@
...
+    - type: redirect
+      matcher: /pouch_api
+      target: http://host.docker.internal:6000
+
+    - type: fishtail_ingest
+      matcher: /pouch_api/ingest
+      target: http://host.docker.internal:6000/api/v1/ingest
+      fishtail_ingest_config:
+        subscribed_collections:
+          - lexicon: app.bsky.feed.like
```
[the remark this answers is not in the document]
eagraf: Yeah there's a lot of interesting stuff that can get specified here in the future.

internal/node/fishtail/firehose.go
```diff
@@ -0,0 +1,151 @@
...
+// LinkedRecordIngester starts a goroutine dedicated to processing ingestion chains as they are
+// received from the FirehoseConsumer. Each chain starts with the initial record that was specified
+// in the firehose event, and then adds any linked records to the chain as they are discovered.
+func (f *Fishtail) LinkedRecordIngester() func() error {
+	return func() error {
+		return f.ingester.Run()
+	}
+}
+
+func (f *Fishtail) runFirehoseConsumer(ctx context.Context, relayHost string) error {
```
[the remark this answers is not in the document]
eagraf: The websocket detail comes directly from atproto / the PDS implementation. No way around it

PR #30, opened 2024-09-22
Support setting up default apps in habitat.yaml

eagraf:
This is a small QOL improvement that saves a lot of time when manually testing the node. Instead of clicking through Postman or another HTTP client to create a new app and start it through the node API, simply add an entry to the `habitat.yml` config file. For example, this configuration can be used to start the Pouch app:

```
default_apps:
  pouch_frontend:
    app_installation:
      name: pouch_frontend
      version: 3
      driver: web

      driver_config:
        download_url: https://github.com/eagraf/extension-save-to-pocket/releases/download/demo-release-2/dist.tar.gz
        bundle_directory_name: pouch

      registry_url_base: a
      registry_app_id: b
      registry_tag: c

    reverse_proxy_rules:
      - type: file
        matcher: /pouch
        target: pouch/3/dist
      - type: redirect
        matcher: /pouch_api
        target: http://100.113.121.9:5000
      - type: fishtail_ingest
        matcher: /pouch_api/ingest
        target: http://100.113.121.9:5000/api/v1/ingest
        fishtail_ingest_config:
          subscribed_collections:
            - lexicon: app.bsky.feed.like
            - lexicon: com.habitat.pouch.link
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,41 @@ make run-dev-fresh
 ```
 You should now see a bunch of logs indicating the node has come up.
 
+### Configuration for Local Development
+In development, the node can be configured using environment variables or through the `habitat.yml` file in the `.habitat` directory at the root of your repository. For a full list of configuration options, check out the `internal/node/config/config.go` file. When developing Habitat apps, it is often helpful to configure the `default_apps` field in the `habitat.yml` file. Doing so will automatically install the app for you and start the app when the node starts, which will save you a lot of clicking around. Here's an example of what setting the `default_apps` field could look like:
+
+```
+default_apps:
+  pouch_frontend:
+    app_installation:
+      name: pouch_frontend
+      version: 3
+      driver: web
+
+      driver_config:
+        download_url: https://github.com/eagraf/extension-save-to-pocket/releases/download/demo-release-2/dist.tar.gz
+        bundle_directory_name: pouch
+
+      registry_url_base: a
+      registry_app_id: b
+      registry_tag: c
+
+    reverse_proxy_rules:
+      - type: file
+        matcher: /pouch
+        target: pouch/3/dist
+      - type: redirect
+        matcher: /pouch_api
+        target: http://100.113.121.9:5000
+      - type: fishtail_ingest
+        matcher: /pouch_api/ingest
+        target: http://100.113.121.9:5000/api/v1/ingest
+        fishtail_ingest_config:
+          subscribed_collections:
+            - lexicon: app.bsky.feed.like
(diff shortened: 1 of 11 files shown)


internal/node/controller/init.go
```diff
@@ -108,10 +109,15 @@ func initTranstitions(nodeConfig *config.NodeConfig) ([]hdb.Transition, error) {
 	}
 
 	pdsAppConfig := generatePDSAppConfig(nodeConfig)
-	defaultApplications := []types.PostAppRequest{
-		pdsAppConfig,
+	defaultApplications := []*types.PostAppRequest{
+		&pdsAppConfig,
```
arushibandi, line 113: have `generatePDSAppConfig` return a pointer?

internal/node/config/config.go
```diff
@@ -290,6 +292,34 @@ func (n *NodeConfig) FrontendDev() bool {
...
+	for _, appConfig := range defaultApps {
+		// Marshal the app config into bytes, and then unmarshal into a PostAppRequest
+		appConfigBytes, err := yaml.Marshal(appConfig)
+		if err != nil {
+			log.Error().Err(err).Interface("app", appConfig).Msg("Failed to marshal default app config")
+			continue
+		}
+		log.Info().Str("app", string(appConfigBytes)).Msg("Default app config")
+
+		var request types.PostAppRequest
```
arushibandi, line 308: can we go directly from string -> types.PostAppRequest?

the `Marshal` and `Unmarshal` seems unnecessary and less efficient

README.md
```diff
@@ -37,6 +37,41 @@ make run-dev-fresh
 ```
 You should now see a bunch of logs indicating the node has come up.
 
+### Configuration for Local Development
+In development, the node can be configured using environment variables or through the `habitat.yml` file in the `.habitat` directory at the root of your repository. For a full list of configuration options, check out the `internal/node/config/config.go` file. When developing Habitat apps, it is often helpful to configure the `default_apps` field in the `habitat.yml` file. Doing so will automatically install the app for you and start the app when the node starts, which will save you a lot of clicking around. Here's an example of what setting the `default_apps` field could look like:
+
+```
```
arushibandi, line 43: we should add tests for this -- a bunch of different `habitat.yaml` files and starting the node with that.

arushibandi (approved):

internal/node/config/config.go
eagraf: I found there's a way to user Viper to unmarshal directly

PR #31, opened 2024-09-22
QOL improvements to testing migrations

eagraf:
When testing our schema migrations, we make sure the list of migrations applied on top of each other results in the exact same result as the `schema.json` file. 

This just ensures that the output showing the diffs between the expected and actual result schemas is printed clearly with proper indentation and no escapes, so it's easy to read, which makes debugging issues with this critical test a lot faster.

### DIFF
--- a/core/state/node/migrations.go
+++ b/core/state/node/migrations.go
@@ -1,6 +1,7 @@
 package node
 
 import (
+	"bytes"
 	"embed"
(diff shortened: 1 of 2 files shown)


arushibandi (approved):
bunch of lint errors that need to be fixed but otherwise looks okay to me!

PR #32, opened 2024-09-22
Eagraf/refactor reverse proxy

eagraf:
Long overdue refactor to the reverse proxy sub-system, which makes it much simpler and easier to understand:
- Get rid of the distinction between Handlers and Rules, which ended up with us having duplicate structs that basically stored the same information, which ended up being very confusing. Instead, we have a new function `getHandlerFromRule` which generates an `http.Handler` implementation based off the rule given.
- The active reverse proxy rules are derived entirely from the node's state. Previously, there were several rules such as the node API rule and frontend rule that were added to the reverse proxy, but were not shown in the app's state, which was very confusing. Now all rules are shown there and it is the source of truth.
- Create a new `AddReverseProxyRuleTransition` which does what it says.
-

### DIFF
--- a/core/state/node/core.go
+++ b/core/state/node/core.go
@@ -46,15 +46,17 @@ type Process struct {
 // The semantics of the target field changes depending on the type. For file servers, it represents the
 // path to the directory to serve files from. For redirects, it represents the URL to redirect to.
 type ReverseProxyRule struct {
-	ID      string `json:"id" yaml:"id"`
-	Type    string `json:"type" yaml:"type"`
-	Matcher string `json:"matcher" yaml:"matcher"`
-	Target  string `json:"target" yaml:"target"`
-	AppID   string `json:"app_id" yaml:"app_id"`
+	ID      string               `json:"id" yaml:"id"`
+	Type    ReverseProxyRuleType `json:"type" yaml:"type"`
+	Matcher string               `json:"matcher" yaml:"matcher"`
+	Target  string               `json:"target" yaml:"target"`
+	AppID   string               `json:"app_id" yaml:"app_id"`
 }
 
+type ReverseProxyRuleType = string
+
 const (
-	ProxyRuleFileServer       = "file"
-	ProxyRuleRedirect         = "redirect"
-	ProxyRuleEmbeddedFrontend = "embedded_frontend"
+	ProxyRuleFileServer       ReverseProxyRuleType = "file"
+	ProxyRuleRedirect         ReverseProxyRuleType = "redirect"
+	ProxyRuleEmbeddedFrontend ReverseProxyRuleType = "embedded_frontend"
 )
--- a/core/state/node/core.go
+++ b/core/state/node/core.go
@@ -52,3 +52,9 @@ type ReverseProxyRule struct {
 	Target  string `json:"target" yaml:"target"`
 	AppID   string `json:"app_id" yaml:"app_id"`
 }
+
+const (
+	ProxyRuleFileServer       = "file"
+	ProxyRuleRedirect         = "redirect"
+	ProxyRuleEmbeddedFrontend = "embedded_frontend"
+)
--- a/internal/node/controller/controller_test.go
+++ b/internal/node/controller/controller_test.go
@@ -27,7 +27,7 @@ func setupNodeDBTest(ctrl *gomock.Controller, t *testing.T) (NodeController, *mo
 	mockedManager.EXPECT().CreateDatabase(gomock.Any(), gomock.Any(), gomock.Any()).DoAndReturn(
 		// signature of anonymous function must have the same number of input and output arguments as the mocked method.
 		func(nodeName, schemaName string, initTransitions []hdb.Transition) (hdb.Client, error) {
-			require.Equal(t, 2, len(initTransitions))
+			require.Equal(t, 4, len(initTransitions))
 
 			initStateTransition := initTransitions[0]
 			require.Equal(t, node.TransitionInitialize, initStateTransition.Type())
--- a/internal/node/controller/controller_test.go
+++ b/internal/node/controller/controller_test.go
@@ -1,7 +1,6 @@
 package controller
 
 import (
-	"crypto/x509"
 	"encoding/json"
 	"errors"
 	"testing"
@@ -42,11 +41,10 @@ func setupNodeDBTest(ctrl *gomock.Controller, t *testing.T) (NodeController, *mo
 			return mockedClient, nil
 		}).Times(1)
 
-	controller, err := NewNodeController(mockedManager, &config.NodeConfig{
-		RootUserCert: &x509.Certificate{
-			Raw: []byte("root_cert"),
-		},
-	})
+	config, err := config.NewTestNodeConfig(nil)
+	require.Nil(t, err)
+
+	controller, err := NewNodeController(mockedManager, config)
 	controller.pdsClient = mockedPDSClient
 	require.Nil(t, err)
 	err = controller.InitializeNodeDB()
--- a/internal/node/controller/init.go
+++ b/internal/node/controller/init.go
@@ -85,9 +85,14 @@ func generateDefaultReverseProxyRules(nodeConfig *config.NodeConfig) ([]*node.Re
 		Matcher: "", // Root matcher
 	}
 	if nodeConfig.FrontendDev() {
+		// In development mode, we run the frontend in a separate docker container with hot-reloading.
+		// As a result, all frontend requests must be forwarde to the frontend container.
 		frontendRule.Type = node.ProxyRuleRedirect
 		frontendRule.Target = "http://habitat_frontend:8000/"
 	} else {
+		// In production mode, we embed the frontend into the node binary. That way, we can serve
+		// the frontend without needing to set it up on the host machine.
+		// TODO @eagraf - evaluate the performance implications of this.
 		frontendRule.Type = node.ProxyRuleEmbeddedFrontend
 	}
 
--- a/internal/node/controller/init.go
+++ b/internal/node/controller/init.go
@@ -2,6 +2,7 @@ package controller
 
 import (
 	"fmt"
+	"net/url"
 	"path/filepath"
 
 	"github.com/docker/docker/api/types/mount"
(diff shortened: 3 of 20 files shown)


core/state/node/core.go
```diff
@@ -52,3 +52,9 @@ type ReverseProxyRule struct {
 	Target  string `json:"target"`
 	AppID   string `json:"app_id"`
 }
+
+const (
```
arushibandi, line 56: wanna make `type ProxyRule string` then type the following?

internal/node/controller/controller_test.go
```diff
@@ -27,7 +27,7 @@ func setupNodeDBTest(ctrl *gomock.Controller, t *testing.T) (NodeController, *mo
 	mockedManager.EXPECT().CreateDatabase(gomock.Any(), gomock.Any(), gomock.Any()).DoAndReturn(
 		// signature of anonymous function must have the same number of input and output arguments as the mocked method.
 		func(nodeName, schemaName string, initTransitions []hdb.Transition) (hdb.Client, error) {
-			require.Equal(t, 2, len(initTransitions))
+			require.Equal(t, 4, len(initTransitions))
```
arushibandi, line 30: should we make this a var equal to something instead of a magic number?

core/state/node/core.go
```diff
@@ -52,3 +52,9 @@ type ReverseProxyRule struct {
...
+	ProxyRuleFileServer       = "file"
+	ProxyRuleRedirect         = "redirect"
```
arushibandi, line 58: also can we move all of these into the `reverse_proxy` package?

so we can do stuff like `reverse_proxy.Rule` instead of `node.ReverseProxyRule`

internal/node/reverse_proxy/restore_test.go
```diff
@@ -72,5 +72,5 @@ func TestProcessRestorer(t *testing.T) {
 	err = restorer.Restore(restoreUpdate)
 	assert.Nil(t, err)
 
-	assert.Equal(t, 1, len(restorer.ruleSet.rules))
+	assert.Equal(t, 2, len(restorer.ruleSet.rules))
```
arushibandi, line 75: was this broken before?

internal/node/reverse_proxy/server.go
```diff
@@ -37,33 +25,49 @@ func NewProxyServer(logger *zerolog.Logger, config *config.NodeConfig) *ProxySer
...
 	// Find the matching rule with the highest "rank", aka the most slashes '/' in the URL path.
 	highestRank := -1
 	for _, rule := range s.RuleSet.rules {
-		if rule.Match(r.URL) {
-			if rule.Rank() > highestRank {
-				bestMatch = rule
-				highestRank = rule.Rank()
+		if rule != nil {
+			if matchRule(r.URL, rule) {
+				rank := rankMatch(r.URL, rule)
```
arushibandi, line 42: when can we get rid of the ranking logic? i thought go had some default stuff in the http package that could do this

internal/node/reverse_proxy/server.go
```diff
@@ -101,3 +105,23 @@ func (s *ProxyServer) Listener(addr string) (net.Listener, error) {
...
+			return true
+		}
+		return strings.HasPrefix(prefixRemoved, "/")
+	}
+	return false
+
+}
+
+// Find the rank of a match, given a requestURL and a rule
+func rankMatch(requestURL *url.URL, rule *node.ReverseProxyRule) int {
```
arushibandi, line 125: :(

internal/node/reverse_proxy/subscriber.go
```diff
@@ -8,6 +8,25 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
+func NewProcessProxyRuleStateUpdateSubscriber(ruleSet *RuleSet) (*hdb.IdempotentStateUpdateSubscriber, error) {
```
arushibandi, line 11: I hate this name 😭 is there a better name or can we at least add a comment?

internal/node/reverse_proxy/subscriber.go
```diff
@@ -47,3 +66,36 @@ func (e *ProcessProxyRulesExecutor) Execute(update hdb.StateUpdate) error {
...
+type AddProxyRulesExecutor struct {
+	RuleSet *RuleSet
+}
+
+func (e *AddProxyRulesExecutor) TransitionType() string {
+	return node.TransitionAddReverseProxyRule
+}
+
+func (e *AddProxyRulesExecutor) ShouldExecute(update hdb.StateUpdate) (bool, error) {
+	// This process is lightweight, so we can execute it every time
```
arushibandi, line 79: hmm, isn't this bad practice though? like if we ever wanna test what executed, it doesn't make sense for this to execute on everything. plus seems like we may have many of these in the future

internal/node/reverse_proxy/proxy.go
```diff
@@ -2,59 +2,28 @@ package reverse_proxy
...
-			return err
-		}
-	} else {
-		return fmt.Errorf("unknown rule type %s", rule.Type)
+	if _, ok := rs.rules[rule.ID]; ok {
+		return fmt.Errorf("reverse proxy rule with id %s is already present", rule.ID)
+	}
+
+	// Make sure the rule type is valid.
+	if rule.Type != node.ProxyRuleRedirect && rule.Type != node.ProxyRuleFileServer && rule.Type != node.ProxyRuleEmbeddedFrontend {
```
arushibandi, line 21: I think if we type the rules properly we won't need this

internal/node/reverse_proxy/handler_test.go
```diff
@@ -0,0 +1,24 @@
...
+		t.Error(err)
+	}
+
+	redirectRule := &node.ReverseProxyRule{
+		ID:     "redirect1",
+		Type:   node.ProxyRuleRedirect,
+		Target: "http://fake-target/api",
+	}
+
+	getHandlerFromRule(redirectRule, fakeConfig)
```
arushibandi, line 23: we're not asserting on anything in this test? +1 no err handler

internal/node/controller/init_test.go
```diff
@@ -0,0 +1,106 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		newState, err := hdb.NewJSONState(schema, newStateBytes)
+		if err != nil {
+			return nil, err
+		}
+
+		oldJSONState = newState
```
arushibandi, line 66: hm instead of doing this can we define
`var newState` before the loop and use that? this seems unnecessarily confusing

internal/node/controller/init_test.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func TestInitTransitions(t *testing.T) {
+	config, err := config.NewNodeConfig()
+	require.Nil(t, err)
+
+	transitions, err := initTranstitions(config)
+	require.Nil(t, err)
+
+	require.Equal(t, 4, len(transitions))
```
arushibandi, line 87: magic number? can we have a test method that counts transitions from a config?

internal/node/controller/init_test.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func TestFrontendDevMode(t *testing.T) {
```
arushibandi, line 90: can we combine this test and the last or not have it re-run a bunch of the same stuff?

internal/node/controller/init.go
```diff
@@ -72,6 +73,34 @@ func generatePDSAppConfig(nodeConfig *config.NodeConfig) types.PostAppRequest {
...
+	apiURL, err := url.Parse(fmt.Sprintf("http://localhost:%s", constants.DefaultPortHabitatAPI))
+	if err != nil {
+		return nil, err
+	}
+
+	frontendRule := &node.ReverseProxyRule{
+		ID:      "default-rule-frontend",
+		Matcher: "", // Root matcher
+	}
+	if nodeConfig.FrontendDev() {
```
arushibandi, line 86: comments? 

i'm a little confused

internal/node/reverse_proxy/handler.go
```diff
@@ -0,0 +1,146 @@
...
+)
+
+func getHandlerFromRule(rule *node.ReverseProxyRule, nodeConfig *config.NodeConfig) (http.Handler, error) {
+	switch rule.Type {
+	case node.ProxyRuleRedirect:
+		return getRedirectHandler(rule)
+	case node.ProxyRuleFileServer:
+		return getFileServerHandler(rule, WithBasePath(nodeConfig.WebBundlePath()))
+
+	case node.ProxyRuleEmbeddedFrontend:
```
arushibandi, line 29: what's the difference between embedded frontend and file server? seems like they're both file servers but embedded frontend is special?

internal/node/reverse_proxy/handler.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+
+		http.FileServer(http.Dir(path)).ServeHTTP(w, r)
+	}
+}
+
+// Options for file server rules
+
+type FileServerOptions struct {
+	EmbeddedFS fs.FS  // Instead of using Path, pass in an fs.FS. Useful for embedding the Habitat frontend.
```
arushibandi, line 130: can you explain a little more what embedding does? i read the go docs but confused why we sometimes need it.

internal/node/controller/controller_test.go
eagraf: I don't think it's a big deal to have magic numbers in tests. If it breaks, the person fixing it will have to stop and think what changed before updating it anyways.

core/state/node/core.go
eagraf: I don't thinks so without creating a bunch of import loops. These were originally in there, but they moved once the reverse proxy rules became a part of the state.

internal/node/reverse_proxy/server.go
eagraf: Do you know where that is? Overall, I think we'll probably always need something like this, especially if there's some dynamic matching fro routes in the future.

internal/node/reverse_proxy/restore_test.go
eagraf: The behavior for restoring changed. Previously, it looked at started processes, and worked backwards to figure out what rules corresponded to them. The new behavior is to just look directly at the reverse proxy rules list directly. Imo, the reverse_proxy_rules state field should be the source of truth, and our transitions for stopping processes should be responsible for deactivating rules by removing them from the list.

The reason I made the change was because there are some default rules that don't correspond to processes which the old logic didn't handle.

internal/node/reverse_proxy/subscriber.go
eagraf: You're right + this won't exactly be lightweight if we have multiple nodes talking over raft.

eagraf: I can reword this, there isn't exactly much to check here. Any conflicts should have been handled by the validation step in the state machine.

internal/node/reverse_proxy/handler.go
eagraf: The frontend isn't sitting on the host machine's file system, instead, when compiling the node binary we embed the entire built frontend bundle into the binary, so that we can serve it directly. This is only used when we aren't in development mode, and is useful because it means there's no setup step where we have to first write out the frontend to a directory on the machine after the initial binary has been installed.

Added some comments explaining this as well

internal/node/controller/init_test.go
eagraf: whatAbout renaming oldJSONState to prevState?

internal/node/controller/init_test.go
eagraf: Same response here, I don't think we should get bogged down creating a constant for every number value in tests

internal/node/reverse_proxy/handler_test.go
eagraf: Nice catch

internal/node/reverse_proxy/subscriber.go
eagraf: I got rid of the StateUpdate part, I think we can ignore that for other ones like this going forward

internal/node/controller/init_test.go
eagraf: There's not that much overlap here, and also these should all run pretty fast anyways

internal/node/reverse_proxy/handler.go
eagraf: Added comments

internal/node/reverse_proxy/proxy.go
eagraf: I tried this, but TIL go will infer the type here

PR #33, opened 2024-11-09
Eagraf/frontend improvements

eagraf:
Screen Shot 2024-11-09 at 12 14 32 AM
Screen Shot 2024-11-09 at 12 14 45 AM
Screen Shot 2024-11-09 at 12 20 04 AM
* Add a "My Server" page to show basic node state including running apps, reverse proxy rules, and an unimplimented user page.
* Add a new "App Gallery" page that lists available apps (currently only visible in dev mode). The available apps are read from an embedded yaml file, that is templated so system specific values can be inserted at runtime.
* The home page has tiles that link to app frontends such as Pouch, as well as the two pages mentioned above.
* Some visual improvements to the header, CSS, system day/night mode settings, etc.
* Typescript types are generated from the Go types of the node using the new Make rule: `frontend/types/api.ts`


.gitignore
```diff
@@ -15,3 +15,5 @@ dev.env
 
 # Ignore frontend build
 internal/frontend/build
+next
+frontend@*
```
arushibandi, line 19: why `@`?

internal/node/controller/routes_test.go
```diff
@@ -228,12 +228,6 @@ func TestGetNodeHandler(t *testing.T) {
 
 	var respBody types.GetNodeResponse
 	require.NoError(t, json.Unmarshal(bytes, &respBody))
-	for k, v := range respBody.State {
```
arushibandi, line 231: why did this get deleted?

cmd/node/main.go
```diff
@@ -137,13 +138,17 @@ func main() {
...
 		api.NewVersionHandler(),
 		controller.NewGetNodeRoute(db.Manager),
 		controller.NewLoginRoute(&controller.PDSClient{NodeConfig: nodeConfig}),
 		controller.NewAddUserRoute(nodeCtrl),
 		controller.NewInstallAppRoute(nodeCtrl),
 		controller.NewStartProcessHandler(nodeCtrl),
 		controller.NewMigrationRoute(nodeCtrl),
+
+		// App store routes
+		appstore.NewAvailableAppsRoute(nodeConfig),
```
arushibandi, line 151: I wonder if we should refactor this so that instead we pass in some function like `registerRoute` which calls `router.Handle` under the hood to all of the down streams, that they can use.

In the long term, it seems annoying to have to add every single route into `main.go`. We probably want to do this for each component inside of it's entrypoint, like in a central place for controller, appstore etc. so that if you wanna update app store, you don't need to update main.go

core/state/node/schema.go
```diff
@@ -44,12 +44,12 @@ const AppStateInstalled = "installed"
 const AppStateUninstalled = "uninstalled"
 
 type AppInstallationState struct {
-	*AppInstallation
-	State string `json:"state"`
+	*AppInstallation `tstype:",extends,required"`
```
arushibandi, line 47: sorry, i've never used `tstype` -- is there documentation you can point me to?

frontend/app/login/page.tsx
```diff
@@ -28,10 +28,11 @@ const LoginInternal = () => {
...
 
     return (
-        <div className="login-container">
-            <form className="login-form" onSubmit={handleSubmit}>
+        <div className={styles.loginBody}>
+            <div className={styles.loginContainer}>
+                <form className={styles.loginForm} onSubmit={handleSubmit}>
                 <h2>Login</h2>
-                <div className="form-group">
+                <div className={styles.formGroup}>
```
arushibandi, line 35: why did these all get changed? did you change the framework you're using?

frontend/types/api.ts
```diff
@@ -0,0 +1,58 @@
...
+  password: string;
+  inviteCode: string;
+}
+export type PDSCreateAccountResponse = { [key: string]: any};
+export interface PDSCreateSessionRequest {
+  identifier: string; // email or handle
+  password: string;
+}
+export type PDSCreateSessionResponse = { [key: string]: any};
+export interface PDSGetRecordRequest {
```
arushibandi, line 49: where do these get generated from? from Go territory?

arushibandi (approved):
i didn't take the closest eye to all the frontend changes, but trust you + the screenshots

.gitignore
eagraf: These are generated by next when building the frontend using the make rules

cmd/node/main.go
eagraf: Yeah I agree, I can note that for the future.

core/state/node/schema.go
eagraf: You can check out the repo: https://github.com/gzuidhof/tygo

This annotation was so it could handle the embedded struct correctly.

frontend/app/login/page.tsx
eagraf: I switched to using css modules for this dir, to isolate the CSS since it was leaking into some other stuff. This is the syntax we were using in other areas with those modules, so I decided to keep it consistent.

internal/node/controller/routes_test.go
eagraf: I don't think this for loop was doing anything real, it's just looping through each key and value, and then accessing those keys in the map and checking the value matches.

frontend/types/api.ts
eagraf: Yup, checkout the changes to the Makefile to see the rules that generate these.

PR #34, opened 2024-11-25
Eagraf/state machine

eagraf:


arushibandi:
changed base so i can see the diff more clearly

arushibandi (commented):
Some small comments, need to take a look at the bulk of it still

internal/node/controller/controller.go
```diff
@@ -135,14 +141,121 @@ func (c *BaseNodeController) FinishAppInstallation(userID string, appID, registr
...
+		}
+	}
+
+	if previouslyRunning {
+
+		err = c.StopProcess(process.ID)
+		if err != nil {
+			return err
+		}
+		err := c.WaitForState(func(updatedState hdb.State) (bool, error) {
```
arushibandi, line 168: hm this feels racy. what if there's a race condition where something else happens (say it crashes) and we never go into the upgrade state? 

I wonder if there's a better way to do this control flow.

internal/node/controller/wait.go
```diff
@@ -0,0 +1,88 @@
...
+		return nil
+	}
+
+	// Add the subscriber to the channel and make sure to remove it when we're done
+	c.stateUpdatesChannel.AddSubscriber(subscriber)
+	// Remove the subscriber when we're done
+	defer c.stateUpdatesChannel.RemoveSubscriber(subscriber)
+
+	// Set up a timeout channel
+	timeoutChan := make(chan struct{})
```
arushibandi, line 74: I think you can make use of `time.After` here https://pkg.go.dev/time#After

Also still thinking about this logic, wondering if there's a better way we can do this whole waiting for state transitions thing.

Curious how it works in other systems, like ecs/k8s. maybe there's some prior stuff we can look at here.

PR #35, opened 2024-12-22
pds: small fixes

arushibandi:
- set `PDS_INVITE_REQUIRED=false` in the default PDS envs to make testing a little easier. we were just bypassing this anyways by doing it all internally. (need to clear volumes to get updated PDS container)
- PDS client shouldn't take in node config, only what it needs. right now this is just admin password.
- reuse pds client rather than creating it in two places -- make this private actually so that this is harder to mess up
- randome comment / log cleanup

### DIFF
--- a/config/air.node.toml
+++ b/config/air.node.toml
@@ -17,7 +17,7 @@ stop_on_error = true
 # When in debugging mode, it sometimes helps to have the application block immediately so that you can
 # easily debug code that executes near startup. To do this, remove the --continue flag and rebuild the docker
 # image.
-full_bin = "dlv exec --accept-multiclient --log --headless --continue --listen :4000 --api-version 2 /habitat_node"
+full_bin = "dlv exec --accept-multiclient --log --headless --listen :4000 --api-version 2 /habitat_node"
 
 # Watch these filename extensions.
 include_ext = ["go", "yml"]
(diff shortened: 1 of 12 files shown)


frontend/types/api.ts
```diff
@@ -46,13 +46,3 @@ export interface PDSCreateSessionRequest {
...
-export interface PDSGetRecordRequest {
-  repo: string;
-  collection: string;
-  rkey: string;
-}
-export interface PDSGetRecordResponse {
-  uri: string;
-  cid: string;
-  value: { [key: string]: any};
-}
```
arushibandi, line 58: this automatically happened -- i'm guessing these got checked in in another PR but we're not actually using them yet? or something?

arushibandi: see the fishtail stuff that got removed below too

config/air.node.toml
```diff
@@ -17,7 +17,7 @@ stop_on_error = true
 # When in debugging mode, it sometimes helps to have the application block immediately so that you can
 # easily debug code that executes near startup. To do this, remove the --continue flag and rebuild the docker
 # image.
-full_bin = "dlv exec --accept-multiclient --log --headless --continue --listen :4000 --api-version 2 /habitat_node"
```
eagraf, line 20: Let's add --continue back

frontend/types/api.ts
eagraf: I think these were originally on the fishtail branch and split off later on, so probably fine

eagraf (approved):

PR #37, opened 2024-12-22
stop plumbing node config everywhere

arushibandi:
Config is generally bad, we should be passing in only the arguments that are needed directly to constructors.
This makes it a lot easier for debugging, and also testing -- we don't need to generate fake node configs everywhere and naming dependences is a lot clearer -- i.e. reverse proxy takes in a default serve path rather than node config and calling config.WebBundlePath().

I know this might not be the most important order of business, but keeping code organized helps me a lot to hit the ground running.


cmd/node/main.go
```diff
@@ -111,7 +119,30 @@ func main() {
...
+	// Generate the list of apps to have installed and started when the node first comes up
+	pdsAppConfig := generatePDSAppConfig(nodeConfig)
+	defaultApps := append([]*types.PostAppRequest{
+		pdsAppConfig,
+	}, nodeConfig.DefaultApps()...)
+	log.Info().Msgf("configDefaultApps: %v", defaultApps)
+
+	initialTransitions, err := initTranstitions(initState, defaultApps, proxyRules)
+	if err != nil {
+		log.Fatal().Err(err).Msg("unable to generate proxy rules")
```
eagraf, line 142: This error message doesn't look like it's talking about the line above it

internal/node/controller/init_test.go
```diff
@@ -1,111 +0,0 @@
-package controller
-
-import (
```
eagraf, line 3: I don't see where these tests got moved to - are they still around?

eagraf (approved):

internal/node/controller/init_test.go
arushibandi: `testTransitions` is a copy of what's in transitions_test.go
TestInitTransitions just tests that there are 4 initial transitions ... which seems useless
the latter tests that the initial transitions include a redirect rule ... which also seems useless>

if we want to be testing something with that, we should probably have some exported defaults and test on those, rather than something that is just used in main.go directly

PR #40, opened 2025-01-12
Refactor `StartProcess` to new way

arushibandi:
Whew, this PR originally started with the goal to remove the restorer/subsriber from the `process` directory in order to make the logic flow more easy to follow and rewrite the flow so that NodeController (or something like it) is a central place of control. I ended up doing a bit more than that as I came across places for optimization and cleanup. Gonna try to write down all the changes here, but I might have missed a couple things:

1. Create a `node.Component[T]` type that can be used to force modules used by the node controller to be able to restore certain state and certain transitions. `SupportedTransitionTypes()` is not actually used anywhere (yet -- I'm considering being a bit stricter about typing at some point), so I'm considering removing it. 
2. Start creating a `CtrlServer`. This will eventually completely replace the `apiServer` we have going, and as a result we can stop needing this server having to implement `Route` and simply serve these routes normally.
3. Start creating a `controller2`. This needs a new name. This is the actual backing for `CtrlServer`, which is just a http server wrapper around this. The goal here is that it will eventually do everything `BaseNodeController` does, but without exporting these functions, because everything that needs to be called out to will happen internally. 
---
After doing that, I realized that there is a bit of redundancy in the node transitions and the different process states.  

1. Specifically, process can appear in the node state as either `Starting` or `Running`, but these are always bound to get out of sync. For e.g. you could mark something as `Running` but then the process crashes. Or it's possible to be in `Running` when it says `Starting`. I replaced both of these with `Started`. This indicates that we would like to be in a running state, but does not distinguish if we have achieved that (since that distinguishment can be incorrect anyways). I believe this gives us the exact same functionality and is easier to reason about. 
2. After doing this I realized the `Stopped` state wasn't actually used anywhere -- existence indicated a `Started` process and non-existence indicated nothing was there. So I removed the states completely to directly reflect this. I had to add a migration for this.
3. Stop enriching `StartProcessTransition`. I have a gut feeling that we don't actually need to be enriching any transitions and should restructure our code to not do this. It's really hard to reason about the state at any point with these `Enrich` functions. I will try to remove each of these and then fully remove that method from the interface. In this case, after I did 1 + 2, we didn't need the extra running step which adds we  `Ext_Driver_ID`, because the node doesn't do anything with it. Conceptually, this is something that should be managed by just the process manager anyways. The one exception to this is that it's possible for the node to crash or something, leaving a process with `Ext_Driver_ID` running, but the opposite is possible too, the node thinks some `Ext_Driver_ID` is running but it's not, or there's many processes we've spun up and the node doesn't know about all of them. Instead, what we should do is to have **drivers** tag processes with the `ID` that we generate in the `StartProcessTransition`, and then have the drivers be able to check if they are still there, and launch them or kill them if needed. I will work on doing this in a follow-up -- we should do this with PDS so habitat doesn't try to launch it every time we re-run even though there's a process already running

---
Follow-up: Add UI for the stop process route.


internal/node/hdb/client.go
```diff
@@ -3,5 +3,6 @@ package hdb
 type Client interface {
 	DatabaseID() string
 	ProposeTransitions(transitions []Transition) (*JSONState, error)
+	ProposeTransitionsEnriched(transitions []Transition) (*JSONState, error)
```
arushibandi, line 6: Need to delete this before merging, was used in an intermediate state.

arushibandi: Done

internal/node/controller/auth.go
```diff
@@ -1,4 +1,4 @@
-package api
+package controller
```
arushibandi, line 1: doing this fixed a import cycle when I was trying to import `api` somewhere else, and also this probably belongs where it's used

internal/node/controller/controller2.go
```diff
@@ -0,0 +1,101 @@
...
+
+	_, err = c.db.ProposeTransitions([]hdb.Transition{
+		transition,
+	})
+	if err != nil {
+		return errors.Wrap(err, "error proposing transition")
+	}
+
+	proc := transition.Process
+	err = c.processManager.StartProcess(proc, app.AppInstallation)
```
eagraf, line 70: It almost feels like the "callback" should be defined in the transition (or at least be 1:1 with it). I think the main reason for this is that we want to be sure that restoring gets us to the same point as where we were when the state was built up organically. But what happens if someone creates a new controller method that also uses `ProcessStartTransition`, but then forgets to call the `processManager.StartProcess` function? Seems like we'd end up in inconsistent state on restore.

That specific example is a bit contrived, but I think it's more likely for things that don't have such similar names. For example, we probably want to hook up the reverse proxy at some point, but you'd need to remember to do that anywhere the `StartProcess` transition is called.

eagraf: Alternatively, we map each transition to `Components` it affects, and they respond automatically to updates.

internal/node/controller/controller2.go
```diff
@@ -0,0 +1,101 @@
...
+	if !ok {
+		return fmt.Errorf("app with ID %s not found", installationID)
+	}
+
+	bytes, err := state.Bytes()
+	if err != nil {
+		return errors.Wrap(err, "error getting state.Bytes()")
+	}
+
+	transition, err := node.GenProcessStartTransition(installationID, bytes)
```
eagraf, line 57: Why do we pass it in as bytes here? Seems like we immediate unmarshal it here: https://github.com/eagraf/habitat-new/pull/40/files#diff-59e76c44f0435924ed3f11fcc8d86189c2ea0d032a5656f2e066ef4dd4d430bdR377

internal/process/manager.go
```diff
@@ -4,73 +4,68 @@ import (
...
+	// ProcessManager should implement Component -- specifically, restore state given a set of processes and apps
+	node.Component[RestoreInfo]
 }
 
-type BaseProcessManager struct {
+type baseProcessManager struct {
 	processDrivers map[string]Driver
-
-	processes map[string]*RunningProcess
+	processes      map[string]*runningProcess
```
eagraf, line 32: My other PR removed these. Is there actually any point in keeping these around anymore?

Ideally components don't have state if we can help it.

internal/node/controller/controller2.go
arushibandi: I think the direction I'm moving in is that the node controller has specific defined methods and those are the only actions you can take on the node via either the UI or CLI tools.

The node controller is the only component that knows that node state even exists and is fully responsible for managing it. We can define the components in such a way to make this easier, but they fundamentally have no idea that the state exists.

I think this makes it way easier to reason about what the state could be and in what way it could be changed. The idea of `Copmonent`s "responding automatically" to updates for e.g. seems like it adds a lot of potential complications without much gain -- any change we want can just be managed by the node controller.

> But what happens if someone creates a new controller method that also uses ProcessStartTransition, but then forgets to call the processManager.StartProcess function?

I'm not really sure if we should be worrying about this right now. I think we should think more in terms of what user actions can be taken. i.e. if you want to hook up the reverse proxy every time you start a process, that's a part of the `/processes/start` endpoint and then it has to live in this function.

In the long term, what we could have is Components implement like 
`TransitionActions map[TransitionType]func ()` and on updates the node calls each of these. But then again you run into ordering problems for example -- how do you know which actions to take first? Do you always have to start the process before registering the reverse proxy? Seems like yes -- so in the end I still think we'd want to be extra explicit about that.

internal/process/manager.go
arushibandi: Agree, maybe I just need to rebase to pick up that change?

internal/node/controller/controller2.go
arushibandi: Oops, will remove that extra work.

core/state/node/migrations/0000000007-processes-cleanup.json
```diff
@@ -0,0 +1,40 @@
+{
+    "old_version": "v0.0.6",
+    "new_version": "v0.0.7",
+    "up": [
+        {
+            "op": "remove",
+            "path": "/properties/processes/additionalProperties/properties/state"
+        },
+        {
+            "op": "remove",
```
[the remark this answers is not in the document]
arushibandi: I was thinking that this can be implemented independently of node state -- i.e. we shouldn't be opinionated about whether there is an external driver id or how the process gets started. The contract between the node controller and the process manager is:
- controller gives you info to start a process + an ID that it will use to refer to it onwards
- process manager starts that process and if it does need to persist any state about it, it will do so independently

I was thinking that for e.g. we can use [docker labels](https://docs.docker.com/reference/dockerfile/#label) and then the process manager can easily get proc ID -> driver ID at any time. Seems like a moot point for the web driver since it doesn't actually do anything right now.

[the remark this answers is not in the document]
arushibandi: I also feel like in the long term, we'll want some API that lets components store state fully independently of node state. Mainly so that we can iterate on components at a different rate that we iterate on the node / node controller especially if other people want to write their own components.

internal/process/manager.go
arushibandi: Hm, I realized I'm using this for tests (the `GetProcess` and `ListProcesses` functions), maybe we can remove it in a follow-up?

eagraf (approved):

PR #41, opened 2025-01-21
Remove external driver ID

arushibandi:
I forgot to remove `Ext_Driver_ID` from the `StopProcess` call.
Also a bit of stronger typing for `node.ProcessID`
Type fixes in the frontend that I missed in the last PR

Added a simple CLI so I can test out this `StopProcess` as well (see photo)

Screenshot 2025-01-20 at 7 26 54 PM


Makefile
```diff
@@ -26,6 +26,10 @@ GOBIN ?= $$(go env GOPATH)/bin
 
 build: $(TOPDIR)/bin/amd64-linux/habitat $(TOPDIR)/bin/amd64-darwin/habitat
 
+build-ctrl:
```
eagraf, line 29: Since the outcome of this build rule is a physical artifact, we can name the rule the path to the compiled bin. This allows the build system to take advantage of pre-built targets if other targets depend on them.

Then, to keep it easy to run it, we can make the `build-ctrl` just a phony shortcut, that always forces a rebuild.

```suggestion

.PHONY: build-ctrl

build-ctrl: clean-node-ctrl $(TOPDIR)/bin/node-ctrl

clean-build-ctrl:
    rm $(TOPDIR)/bin/node-ctrl

$(TOPDIR)/bin/node-ctrl:
     go build -o $(TOPDIR)/bin/node-ctrl $(TOPDIR)/cmd/node_ctrl/main.go
     
 clean:: clean-build-ctrl
```
The last line makes it so `make clean` includes cleaning up the CLI as well.

Makefile
```diff
@@ -26,6 +26,10 @@ GOBIN ?= $$(go env GOPATH)/bin
 
 build: $(TOPDIR)/bin/amd64-linux/habitat $(TOPDIR)/bin/amd64-darwin/habitat
```
eagraf, line 27: We can add the rule for building the CLI here as another pre-requisite, so everything get's rebuilt at once

Makefile
```diff
@@ -112,14 +116,14 @@ $(TOPDIR)/bin: $(TOPDIR)
 
 # Linux AMD64 Builds
 $(TOPDIR)/bin/amd64-linux/habitat: $(TOPDIR)/bin frontend_server/build
-	GOARCH=amd64 GOOS=linux go build -o $(TOPDIR)/bin/amd64-linux/habitat $(TOPDIR)/cmd/node/main.go
+	GOARCH=amd64 GOOS=linux go build -o $(TOPDIR)/bin/amd64-linux/habitat $(TOPDIR)/cmd/node/main.go && go build -o $(TOPDIR)/bin/amd64-linux/node-ctrl $(TOPDIR)/cmd/node_ctrl/main.go
```
eagraf, line 119: This targetis a file path, so it's main artifact should just be the bin compiled to `$(TOPDIR)/bin/amd64-linux/habitat`. IIRC currently this rule is only used by the release script. If you want to include the CLI in the release, we can just add the target for the CLI bin as a prerequisite to the main `build` rule.

cmd/node_ctrl/main.go
```diff
@@ -0,0 +1,137 @@
...
+				return err
+			}
+			return printResponse(res)
+		},
+	}
+}
+
+func main() {
+	commands := map[string]*cli.Command{
+		"start-process":  startProcess(),
```
eagraf, line 103: Is it easy to make subcommands for `process start`, `process stop` and `process list` with this CLI library? This sort of thing helps a lot with possible future CLI features like help flags, tab-complete, etc.

cmd/node_ctrl/main.go
```diff
@@ -0,0 +1,137 @@
...
+
+var port string
+
+func printResponse(res *http.Response) error {
+	slurp, err := io.ReadAll(res.Body)
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("Status: %s\nResponse: %s\n", res.Status, string(slurp))
```
eagraf, line 26: It'd be really nice if we just output the raw JSON, that way this would work with tools like `jq` out of the box.  We could either add a flag to print just JSON output, or wrap the response JSON with a parent JSON that has the response status. Also, any non-200 response status should reflect in an exit code of 1.

internal/node/controller/server.go
```diff
@@ -69,7 +72,30 @@ func (s *CtrlServer) StopProcess(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
+}
+
+func (s *CtrlServer) ListProcesses(w http.ResponseWriter, r *http.Request) {
+	fmt.Println("hi")
```
eagraf, line 85: Leftover print

internal/docker/driver.go
```diff
@@ -32,47 +38,66 @@ func (d *dockerDriver) Type() string {
...
 	for _, port := range dockerConfig.ExposedPorts {
 		exposedPorts[nat.Port(port)] = struct{}{}
 	}
 
-	createResp, err := d.client.ContainerCreate(context.Background(), &container.Config{
+	createResp, err := d.client.ContainerCreate(ctx, &container.Config{
 		Image:        fmt.Sprintf("%s/%s:%s", app.RegistryURLBase, app.RegistryPackageID, app.RegistryPackageTag),
 		ExposedPorts: exposedPorts,
 		Env:          dockerConfig.Env,
+		Labels: map[string]string{
```
eagraf, line 62: Nice!

internal/docker/driver.go
```diff
@@ -14,6 +15,11 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
+const (
+	habitatLabel   = "habitat_proc_id"
+	errNoProcFound = "no process found with id %s"
```
eagraf, line 20: Maybe "no container found with label" is more direct?

internal/docker/driver.go
```diff
@@ -32,47 +38,66 @@ func (d *dockerDriver) Type() string {
...
+	log.Info().Msgf("Started docker container %s", createResp.ID)
 
-	return createResp.ID, nil
+	return nil
 }
 
-func (d *dockerDriver) StopProcess(extProcessID string) error {
-	err := d.client.ContainerStop(context.Background(), extProcessID, container.StopOptions{})
+func (d *dockerDriver) StopProcess(ctx context.Context, processID node.ProcessID) error {
+	ctrs, err := d.client.ContainerList(ctx, container.ListOptions{})
```
eagraf, line 84: The `ListOptions` struct probably has a way to filter by labels. It probably makes more sense to use docker's implementation here.

eagraf (commented):
Awesome idea with the labeling

cmd/node_ctrl/main.go
arushibandi: I think it's useful to see non-200 response + body in case we decide to return some info in the result.
Non-200 doesn't necessarily mean an error in the CLI program, it could be the right response to return given certain inputs

Makefile
arushibandi: Did this!

Makefile
arushibandi: Did this!

PR #42, opened 2025-01-26
Make process manager more robust

arushibandi:
The process manager was a bit broken because the running processes map got reset every time habitat got restarted, but it's possible that there were still processes running in docker. So for example,
`ListProcesses` and `GetProcess` wouldn't report those processes so the CLI tool was broken.

In general, the actual state of processes can drift from both what's in the node state and whatever is in our in-memory state in the process manager. For the latter, it's best to remove this state and pass all API calls through to the process driver which is backed by docker / something that can check the actual state in the moment. That's what this PR does.

For the former issue, we need to be careful about when we commit transitions or not. This is the code in `controller2.go` and we should be careful about writing a transition if the corresponding action failed. I handled a couple edge cases here -- the process is already started or already stopped -- there might be a couple more but I think it's good enough for now

Also make `type Driver string` so we don't have `"docker"` everywhere


internal/docker/driver.go
```diff
@@ -80,26 +79,55 @@ func (d *dockerDriver) StartProcess(ctx context.Context, process *node.Process,
...
 		),
 	})
 	if err != nil {
-		return err
+		return types.Container{}, false, err
 	}
 
 	if len(ctrs) > 1 {
-		return fmt.Errorf("Got multiple processes with label %s: %v", habitatLabel, ctrs)
+		return types.Container{}, false, fmt.Errorf("Got multiple processes with label=%s: %v", labelVal, ctrs)
```
eagraf, line 94: Nit

```suggestion
		return types.Container{}, false, fmt.Errorf("Got multiple containers with label=%s: %v", labelVal, ctrs)
```

internal/process/manager.go
```diff
@@ -2,103 +2,115 @@ package process
...
 	node.Component[RestoreInfo]
 }
 
+var (
+	ErrDriverNotFound = errors.New("no driver found")
+)
+
 type baseProcessManager struct {
-	processDrivers map[string]Driver
-	processes      map[node.ProcessID]*runningProcess
```
eagraf, line 32: Nice - this should help us avoid a lot of bugs

internal/process/manager.go
```diff
@@ -2,103 +2,115 @@ package process
...
+	return allProcs, nil
 }
 
-func (pm *baseProcessManager) GetProcess(processID node.ProcessID) (*node.Process, bool) {
-	proc, ok := pm.processes[processID]
-	if !ok {
-		return nil, false
+func (pm *baseProcessManager) IsRunning(ctx context.Context, id node.ProcessID) (bool, node.Driver, error) {
+	var derr error
+	for _, driver := range pm.drivers {
```
eagraf, line 65: We should have the process ID include a prefix for the driver type, so we don't have to loop through all the unrelated drivers each time. Especially since some of the drivers will call out to external processes.

internal/process/manager.go
```diff
@@ -2,103 +2,115 @@ package process
...
+		ok, err := driver.IsRunning(ctx, id)
+		if ok && err == nil {
+			fmt.Println("found!", ok, driver.Type())
```
eagraf, line 68: Leftover print

internal/process/manager.go
```diff
@@ -2,103 +2,115 @@ package process
...
 	}
 
-	pm.processes[process.ID] = &runningProcess{
-		Process: process,
-	}
-	return nil
+	return driver.StartProcess(ctx, id, app)
 }
 
+// This could be a bit more efficient if the process manager new which driver the process belonged to
```
eagraf, line 97:
```suggestion
// This could be a bit more efficient if the process manager knew which driver the process belonged to
```

internal/process/manager.go
```diff
@@ -2,103 +2,115 @@ package process
...
+// However, it's possible to get into a state where either the ndoe state does not contain the process and it is running
```
eagraf, line 98:
```suggestion
// However, it's possible to get into a state where either the node state does not contain the process and it is running
```

internal/process/manager.go
eagraf: Same thing here, driver process id prefix would fix this.

internal/node/controller/controller2.go
```diff
@@ -64,38 +65,55 @@ func (c *controller2) startProcess(installationID string) error {
...
-		&node.ProcessStopTransition{
-			ProcessID: processID,
-		},
-	})
+	state, err := c.getNodeState()
 	if err != nil {
 		return err
+	} else if _, ok := state.Processes[processID]; ok {
+		// Only propose transitions if the process exists in state
+		_, err = c.db.ProposeTransitions([]hdb.Transition{
```
eagraf, line 96: This should be redundant with the transition validation right?

internal/node/controller/controller2.go
```diff
@@ -64,38 +65,55 @@ func (c *controller2) startProcess(installationID string) error {
 		return errors.Wrap(err, "error proposing transition")
 	}
 
-	proc := transition.Process
-	err = c.processManager.StartProcess(c.ctx, proc, app.AppInstallation)
+	err = c.processManager.StartProcess(c.ctx, transition.Process.ID, app.AppInstallation)
 	if err != nil {
+		// Rollback the state change if the process start failed
```
eagraf, line 70: Overall, I think this is a good reason for why we'd need `starting` -> `running`-> `stopping` -> `stopped` 
  (and the same for app installations. With these extra transition steps, we can handle the case where an external dependency errors, without the state machine reflecting incorrect information. For example, if we start a process, but Docker fails, we will still report `starting`, which doesn't give anyone else the impression that the app is available. When we detect an error, we can transition to the `errored` state, and further resolution steps could be taken. In general, I feel that having rollback code right next to various attempted transitions is a smell.

In addition, a lot of the potential common causes for `start` transitions to fail will also lead to the `stop` transition in the rollback failing (mainly `docker` being busted / not found).

internal/node/controller/controller2.go
arushibandi: Removed

internal/node/controller/controller2.go
arushibandi: Agree, I think we should add it back in a follow-up. I'll take this on after the other refactors

eagraf (approved):

PR #43, opened 2025-01-29
package manager + reverse proxy refactor

arushibandi:
Mostly deletions. 
Pub-Sub is now only being used for the state update logger.
Some additional cleanup around setup in `main.go` that was racy after some of my previous changes.
Next PR: remove the old controller, delete raft.


internal/node/hdb/state/state_machine.go
```diff
@@ -59,48 +60,48 @@ func NewStateMachine(databaseID string, schema hdb.Schema, initRawState []byte,
 	}, nil
 }
 
-func (sm *StateMachine) StartListening() {
-	go func() {
```
arushibandi, line 63: remove erroneous go func()

core/state/node/migrations/0000000001-init.json
```diff
@@ -80,7 +80,7 @@
                         }
                     }
                 },
-                "additionalProperties": false,
+                "additionalProperties": true,
```
eagraf, line 83: What was this blocking before?

eagraf (approved):
Looks straightforward

core/state/node/migrations/0000000001-init.json
arushibandi: trying to change something from a pointer to not a pointer in the underlying golang struct

PR #44, opened 2025-03-30
Abandi/remove raft

arushibandi:
Remove raft, instead simply writing state to json file. Also removed all the `StateMachine` related stuff -- for now a simple mutex to prevent concurrency will suffice.

TODO:
- there's a bit of an import cycle issue right now where db depends on `node.NodeSchema` to create `hdb.JSONState`, but `node.NodeSchema` depends on `hdb.State`. Need to sort this out. This seems to have gotten around by creating an `hdbms` package, but has code smells, so trying to untangle it. (Fixed in #118)

Test Plan:
- Remove hdb state files and make sure habitat starts up + shows reverse proxy rules  (or check /node/state as expected)
- Copy state into new file path and make sure habitat restarts from existing state + shows reverse proxy rules


internal/node/hdb/state/state_machine.go
```diff
@@ -67,36 +54,20 @@ func (sm *StateMachine) StartListening(ctx context.Context) {
...
 		// TODO this bit of code should be well tested
 		case stateUpdate := <-sm.updateChan:
+			// execute state update
+			stateBytes, err := stateUpdate.NewState().Bytes()
+			if err != nil {
+				log.Error().Err(err).Msgf("error getting new state bytes from state update chan")
+			}
+			jsonState, err := hdb.NewJSONState(sm.schema, stateBytes)
+			if err != nil {
+				log.Error().Err(err).Msgf("error getting new state from state update chan")
```
arushibandi, line 64: this really shouldn't be a log + move on error, it should return something to the caller
should go away with state machine removal

eagraf (approved):

PR #46, opened 2025-04-04
[Permissions] Basic permissions store

sashankg:
https://casbin.org/

Adds a basic (read only for now) permission store that supports:
- read vs write permissions
- roles and inheritance 
  - ex: `did:1` is a `friend` and `friend` has permissions for `app.bsky.post.post1` so `did:1` has permission for `app.bsky.post.pos1`
- glob permissions for nsids 
  - ex: `app.bsky.*` would match the permission for any bsky record
- allow vs deny
  - ex: `did:1` can have `allow` for `app.bsky.*` but `deny` for `app.bsky.post.*` so they could see all bsky records except posts

Tried to explain each line in the model file itself.


core/permissions/permission_store.go
```diff
@@ -0,0 +1,57 @@
...
+	_ "embed"
+
+	"github.com/casbin/casbin/v2"
+	"github.com/casbin/casbin/v2/model"
+	"github.com/casbin/casbin/v2/persist"
+)
+
+type PermissionStore interface {
+	HasPermission(
+		didstr string,
```
eagraf, line 13: IIRC indigo has types for all of these. Not sure if we want to use them or not

eagraf (approved):

core/permissions/permission_store.go
arushibandi: yeah we should use official types for these and i also wonder if instead of `write bool` we should have `permission permissionType` and that can be `write | read` and potentially more ?

arushibandi (approved):

PR #47, opened 2025-04-04
Add privy: a way to store private data on the pds

arushibandi:
Just using NoopEncrypter for now.


internal/node/controller/controller2.go
```diff
@@ -197,3 +215,130 @@ func (c *Controller2) restore(state *node.State) error {
...
+// the shape of the lexicon is { "cid": <cid pointing to the encrypted blob> }
+
+// putRecord with encryption wrapper around this
+func (c *Controller2) putRecord(ctx context.Context, input *agnostic.RepoPutRecord_Input, encrypt bool) (*agnostic.RepoPutRecord_Output, error) {
+	// Not encrypted -- blindly forward the request to PDS
+	if !encrypt {
+		return agnostic.RepoPutRecord(ctx, c.xrpc, input)
+	}
+
+	// Check if a record under this collection already exists publicly with this rkey
```
sashankg, line 235: is this todo? also we need need to check if the rkey exists *privately*, right?

internal/node/api/router.go
```diff
@@ -53,3 +53,28 @@ func NewRouter(
 
 	return routerWithMiddleWare
 }
+
+// Helper package to easily retur
```
eagraf, line 57: Cut off?

internal/privy/privy.go
```diff
@@ -0,0 +1,213 @@
...
+	"strings"
+
+	"github.com/bluesky-social/indigo/api/agnostic"
+	"github.com/bluesky-social/indigo/api/atproto"
+	"github.com/bluesky-social/indigo/atproto/data"
+	"github.com/bluesky-social/indigo/lex/util"
+	"github.com/bluesky-social/indigo/xrpc"
+)
+
+// Privy is an ATProto PDS Wrapper which allos for storing & getting private data.
```
eagraf, line 17:
```suggestion
// Privy is an ATProto PDS Wrapper which allows for storing & getting private data.
```

internal/privy/privy.go
```diff
@@ -0,0 +1,213 @@
...
+// It does this by encrypting data, then storing it in blob. A special lexicon for this purpose,
+// identified by com.habitat.encryptedRecord, points to the blob storing the actual data.
```
eagraf, line 19: Should we get an actual domain to put our lexicons under? I've been using a personal domain I control just so "it's correct".

internal/privy/server.go
```diff
@@ -0,0 +1,116 @@
...
+		}
+	}
+}
+
+// This creates the xrpc.Client to use in the inner privy requests
+// TODO: this should actually pull out the requested did from the url param or input and re-direct there. (Potentially move this lower into the fns themselves).
+// This would allow for requesting to any pds through these routes, not just the one tied to this habitat node.
+func (s *Server) pdsAuthMiddleware(next func(c *xrpc.Client) http.HandlerFunc) http.HandlerFunc {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		bearer := r.Header.Get("Authorization")[7:]
```
eagraf, line 100: Let's do a little validation that `Bearer` is in front just to make sure it's well formatted, so we can give feedback to the client on why this is malformed.

internal/privy/privy.go
```diff
@@ -0,0 +1,213 @@
...
+
+	// If the record with the given collection + rkey identifier was not found (case 2c), attempt to get a private record with permissions look up.
+	if strings.Contains(err.Error(), "RecordNotFound") || strings.Contains(err.Error(), "Could not locate record") {
+		return p.getEncryptedRecord(ctx, xrpc, cid, collection, did, rkey)
+	}
+	// Otherwise the lookup failed in some other way, return the error
+	return nil, err
+}
+
+// getEncryptedRecord assumes that the record given by the cid + collection + rkey + did has been encrypted via putRecord and fetches it
```
eagraf, line 120: Do we really need both cid and rkey to identify the record? Shouldn't rkey be enough?

internal/privy/privy.go
```diff
@@ -0,0 +1,213 @@
...
+	// Record not found results in an error, as does any other non-200 response from the endpoint.
+	// Cases 1a - 1c are handled directly by this case.
+	output, err := agnostic.RepoGetRecord(ctx, xrpc, cid, collection, did, rkey)
+	// If this is a cid lookup (cases 1a-1c) or the record was found (2a + 2b), simply return ()
+	if err == nil {
+		return output, nil
+	}
+
+	// If the record with the given collection + rkey identifier was not found (case 2c), attempt to get a private record with permissions look up.
+	if strings.Contains(err.Error(), "RecordNotFound") || strings.Contains(err.Error(), "Could not locate record") {
```
eagraf, line 113: Seems like we should be able to look at the status code in the error if there was an http error: https://github.com/bluesky-social/indigo/blob/bfc7f487e4ed5ece2c07a1bb309ba1d32dede33e/xrpc/xrpc.go#L84

This is still dependent on the internal implementation of Indigo, but it should be a lot safter than matching strings.

internal/privy/privy.go
```diff
@@ -0,0 +1,213 @@
...
+
+type encryptedRecord struct {
+	Data util.BlobSchema `json:"data" cborgen:"data"`
+}
+
+var (
+	ErrPublicRecordExists               = fmt.Errorf("a public record exists with the same key")
+	ErrNoPutsOnEncryptedRecord          = fmt.Errorf("directly put-ting to this lexicon is not valid")
+	ErrNoEncryptedGetsOnEncryptedRecord = fmt.Errorf("calling getEncryptedRecord on a com.habitat.encryptedRecord is not supported")
+	ErrEncryptedRecordNilValue          = fmt.Errorf("a com.habitat.encryptedRecord record was found but it has a nil value")
```
eagraf, line 43: Let's format in `encryptedRecordNSID` so we don't have to ctrl-F in the future

internal/privy/server.go
```diff
@@ -0,0 +1,116 @@
...
+				AccessJwt: bearer,
+			},
+		}
+		next(c).ServeHTTP(w, r)
+	})
+}
+
+func (s *Server) GetRoutes() []api.Route {
+	return []api.Route{
+		api.NewBasicRoute(http.MethodPost, "/xrpc/com.habitat.putRecord", s.pdsAuthMiddleware(s.PutRecord)),
```
eagraf, line 113: Let's add these as reverse proxy rules on the PDS default app. Ex: https://github.com/eagraf/habitat-new/pull/45/files#diff-41988d47676978b9c3637b55fba6588aed77dab7cff1aadfb41b119964785392R336

(Note in the example, it's actually redirecting something from the oauth stuff to the base PDS, but we'd do the opposite in this case: redirecting from the PDS to our own shim instead).

internal/privy/encrypter.go
```diff
@@ -0,0 +1,62 @@
+package privy
+
+import (
+	"crypto/rand"
+	"crypto/sha256"
+	"encoding/hex"
+
+	"github.com/agl/gcmsiv"
```
eagraf, line 8: Since it's unused, let's comment out this file so we're not adding a new dependency.

arushibandi: okiee it was used in private_data_test.go which is good to make sure we're going back and forth the right way but we can add it back in later

internal/privy/privy.go
arushibandi: sadly because I call `agnostic.RepoGetRecord` I don't have access to the http status code itself. Agree we shouldn't do string matching. I think the proper way would be if they define an error I could do `errors.Is` on or if we make the http request ourselves, also an option.

eagraf (approved):

PR #49, opened 2025-04-05
Eagraf/bffauth

eagraf:
Some auth stuff


internal/node/controller/auth.go
```diff
@@ -103,3 +109,186 @@ func getUserCertAndInfo(controller NodeController, username string) (*x509.Certi
...
+	return &BFFProvider{
+		challengePersister: challengePersister,
+		friends:            friends,
+		signingKey:         signingKey,
+	}
+}
+
+func (p *BFFProvider) GetRoutes() []api.Route {
+	return []api.Route{
+		newRoute(http.MethodPost, "/node/bff/challenge", p.handleChallenge),
```
arushibandi, line 147: should we stop prefixing everything with node? it's all node at this point

internal/node/controller/auth.go
```diff
@@ -103,3 +109,186 @@ func getUserCertAndInfo(controller NodeController, username string) (*x509.Certi
...
+		newRoute(http.MethodPost, "/node/bff/auth", p.handleAuth),
+		newRoute(http.MethodPost, "/node/bff/add_friend", p.handleAddFriend),
+		newRoute(http.MethodGet, "/node/bff/test", p.handleTest),
+	}
+}
+
+func (p *BFFProvider) handleChallenge(w http.ResponseWriter, r *http.Request) {
```
arushibandi, line 154: nit: probably good to leave at least a bit of comments on all of these functions

internal/node/controller/auth.go
```diff
@@ -103,3 +109,186 @@ func getUserCertAndInfo(controller NodeController, username string) (*x509.Certi
...
+	var req BFFChallengeRequest
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		http.Error(w, "Invalid request body", http.StatusBadRequest)
+		return
+	}
```
arushibandi, line 159: we should standardize on if we're using `json.NewDecoder` vs. `json.Unmarshal` ... i prefer the latter if we're not expecting the input to be a huge stream

internal/node/controller/auth.go
```diff
@@ -103,3 +109,186 @@ func getUserCertAndInfo(controller NodeController, username string) (*x509.Certi
...
+	err = json.NewEncoder(w).Encode(map[string]string{
+		"token": token,
+	})
+	if err != nil {
+		http.Error(w, "Failed to encode response", http.StatusInternalServerError)
+		return
+	}
+}
+
+func (p *BFFProvider) handleAddFriend(w http.ResponseWriter, r *http.Request) {
```
arushibandi, line 257: can you explain in a little more detail why we need to do this first (i guess i'm looking for some minimal documentation on the protocol)

internal/node/controller/auth.go
```diff
@@ -103,3 +109,186 @@ func getUserCertAndInfo(controller NodeController, username string) (*x509.Certi
...
+
+	challenge, err := bffauth.GenerateChallenge()
+	if err != nil {
+		http.Error(w, "Failed to generate challenge", http.StatusInternalServerError)
+		return
+	}
+
+	// Create session with UUID
+	sessionID := uuid.New().String()
+	session := &bffauth.ChallengeSession{
```
arushibandi, line 175: can you add a todo somewhere that we shouldn't have multiple per did

internal/node/controller/auth.go
```diff
@@ -103,3 +109,186 @@ func getUserCertAndInfo(controller NodeController, username string) (*x509.Certi
...
+		ExpiresAt: time.Now().Add(15 * time.Minute),
+	}
+	err = p.challengePersister.SaveSession(session)
+	if err != nil {
+		http.Error(w, "Failed to save session", http.StatusInternalServerError)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	err = json.NewEncoder(w).Encode(map[string]string{
```
arushibandi, line 188: same thing here I think `Encoder` is meant for strings -- can we just do `json.Marshal`

internal/node/controller/auth.go
```diff
@@ -103,3 +109,186 @@ func getUserCertAndInfo(controller NodeController, username string) (*x509.Certi
...
+		"session":   sessionID,
+	})
+	if err != nil {
+		http.Error(w, "Failed to encode response", http.StatusInternalServerError)
+		return
+	}
+}
+
+func (p *BFFProvider) handleAuth(w http.ResponseWriter, r *http.Request) {
+	var req struct {
```
arushibandi, line 199: i'd probably just define this as a type.

internal/node/controller/auth.go
```diff
@@ -103,3 +109,186 @@ func getUserCertAndInfo(controller NodeController, username string) (*x509.Certi
...
+	if err != nil {
+		http.Error(w, "Invalid session", http.StatusBadRequest)
+		return
+	}
+
+	err = p.challengePersister.DeleteSession(req.SessionID)
+	if err != nil {
+		http.Error(w, "Failed to delete session", http.StatusInternalServerError)
+		return
+	}
```
arushibandi, line 218: should we have a TODO to not delete it immediately? like we probably want to delete it when it's verified OR have an expiry on it right? (otherwise someone else could come in and end your challenge session)

pkg/bffauth/challenge_test.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"testing"
+
+	"github.com/bluesky-social/indigo/atproto/crypto"
+)
+
+func TestChallengeProofFlow(t *testing.T) {
+	// Generate a test key pair
+	privateKey, err := crypto.GeneratePrivateKeyP256()
+	if err != nil {
```
arushibandi, line 12: same here let's standardize testing
I would go with `require.NoError(t, err, "failed to generate key pair")` here -- require does the same as fatal and then you don't need `err != nil`. i think this is one of the standard packages.

internal/node/controller/auth.go
```diff
@@ -103,3 +109,186 @@ func getUserCertAndInfo(controller NodeController, username string) (*x509.Certi
 
 	return cert, user, nil
 }
+
```
arushibandi, line 112: No reason this needs to be in controller actually ... could just be in `internal/bff`

pkg/bffauth/session.go
```diff
@@ -0,0 +1,53 @@
+package bffauth
+
+import (
+	"fmt"
+	"time"
+)
+
+type ChallengeSessionPersister interface {
```
arushibandi, line 8: I think this should all be internal ... and then we move the specific protocol related pieces out (none of the persisters) into pkg if we want to share with people

arushibandi:
i moved stuff around and addressed some of my nit comments -- the `provider.go` file is below threshold -- i think it might be worth actually testing this (eventually at least), there's some non-trivial stuff in there

internal/node/controller/auth.go
arushibandi: we can make it an internal type if you don't want it to leak out of here

arushibandi: I made this change

internal/node/controller/auth.go
arushibandi: I made this change

pkg/bffauth/session.go
arushibandi: I changed it to this .. then realized it's possible we'd want to persist this beyond process lifetimes -- but I think we'd still want it to be internal and instead have the constructor take in like "bffAuthPersistedStateFile" -- i.e. the caller should be unaware of implementation details like the fact that this is being persisted.

internal/node/controller/auth.go
arushibandi: It's fine if you don't do in this PR, I'm going to take a pass and standardize all of these. Same with our testing.

internal/node/controller/auth.go
arushibandi: Fine to not do in this PR, I'll standardize this as a follow up

arushibandi (approved):

PR #55, opened 2025-04-06
Abandi/get bffs

arushibandi:


internal/bffauth/client.go
```diff
@@ -8,13 +8,21 @@ type ExternalHabitatUser struct {
 	Host      string
 }
 
-type Client struct {
+type Client interface {
+	GetToken(did string) (string, error)
+}
+
+type client struct {
```
sashankg, line 15: nit: I like `clientImpl`

internal/privy/privy.go
```diff
@@ -142,6 +147,13 @@ func (p *store) getEncryptedRecord(ctx context.Context, xrpc *xrpc.Client, cid s
 
 	// Run permissions before returning to the user
 	// if HasAccess(did, collection, rkey) { .... }
+	authz, err := p.permissions.HasPermission(did, collection, rkey, false)
```
sashankg, line 150:
```suggestion
	authz, err := p.permissions.HasPermission(callerDID, collection, rkey, false)
```
?

internal/privy/server.go
```diff
@@ -73,11 +115,60 @@ func (s *Server) GetRecord(cli *xrpc.Client) http.HandlerFunc {
 
 		cid := u.Query().Get("cid")
 		collection := u.Query().Get("collection")
-		did := u.Query().Get("did")
+		repo := u.Query().Get("repo")
 		rkey := u.Query().Get("rkey")
 
-		out, err := s.inner.getRecord(r.Context(), cli, cid, collection, did, rkey)
+		// Try handling both handles and dids
+		id, err := s.dir.LookupHandle(r.Context(), syntax.Handle(repo))
```
sashankg, line 110: is this how they handle `repo` internally? wondering if they already have a method to resolve either case

internal/privy/server.go
```diff
@@ -19,20 +25,33 @@ type PutRecordRequest struct {
...
-		pdsHost: pdsHost,
 		inner: &store{
-			e: enc,
+			e:           enc,
+			permissions: permStore,
 		},
+		habitatResolver: habitatResolver,
+		localPDSHost:    localPDSHost,
+		bffClient:       bffClient,
+		bffServer:       bffServer,
```
sashankg, line 49: looks like we're not passing in a `dir`

internal/privy/server.go
```diff
@@ -73,11 +115,60 @@ func (s *Server) GetRecord(cli *xrpc.Client) http.HandlerFunc {
...
+			Host: s.localPDSHost,
+		}
+
+		targetDID := id.DID.String()
+		callerDID := "" /* unpopulated when unknown */
+		var out *agnostic.RepoGetRecord_Output
+		// If trying to get data from a PDS not managed by habitat
+		if id.PDSEndpoint() != s.localPDSHost {
+			// get bff token
+			token, err := s.bffClient.GetToken(targetDID)
```
sashankg, line 130: fine for now but ideally our our xrpc client would handle this on its own

internal/privy/server.go
```diff
@@ -73,11 +115,60 @@ func (s *Server) GetRecord(cli *xrpc.Client) http.HandlerFunc {
...
+		id, err := s.dir.LookupHandle(r.Context(), syntax.Handle(repo))
 		if err != nil {
+			id, err = s.dir.LookupDID(r.Context(), syntax.DID(repo))
+			if err != nil {
+				http.Error(w, err.Error(), http.StatusBadRequest)
+				return
+			}
+		}
+
+		cli := &xrpc.Client{
```
sashankg, line 119: nit: can we move this closer to where it's used

internal/bffauth/client.go
arushibandi: I think ethan is changing this so i'm going to leave it

internal/privy/privy.go
arushibandi: good catch

sashankg (approved):

PR #57, opened 2025-04-06
Eagraf/bffclient

eagraf:


internal/bffauth/client.go
```diff
@@ -0,0 +1,202 @@
...
+	return map[string]crypto.PrivateKey{
+			"did:plc:avkzrsj7u4kmoq33gx6v4lh4": testKey,
+		}, map[string]*ExternalHabitatUser{
+			"did:plc:avkzrsj7u4kmoq33gx6v4lh4": testUser,
+		}, nil
+
+}
+
+type Client struct {
+	// This is the store of temporary friends that are added to the BFF.
```
arushibandi, line 46: this should actually be an `interface` (see https://github.com/eagraf/habitat-new/pull/55/files#diff-3fcf2c4d8538ad0e112b07632d17fd396d921e24861c08a811370b57277ad8abR11-R13) so it can be mocked out

internal/bffauth/client.go
```diff
@@ -0,0 +1,202 @@
...
+	tempFriendStore map[string]*ExternalHabitatUser
+
+	clientDID string
+
+	privateKey crypto.PrivateKey
+
+	activeTokens map[string]string
+}
+
+func NewClient(clientDID string) (*Client, error) {
```
arushibandi, line 57: Let's replace `tempFriendStore` with a HabitatResolver and a PublicKeyResolver 
So somewhere in internal maybe in `internal/types?` we have

```
type HabitatResolver func(did string) (habitatHost string, err error)
type PublicKeyResolver func(did string) (crypto.PublicKey, err error)
```

Then your `NewClient` can take these in and you can create some mock implementations for testing only. That will get rid of `tempFriendStore`

internal/bffauth/client.go
```diff
@@ -0,0 +1,217 @@
...
+	if err != nil {
+		return "", "", err
+	}
+
+	return respBody.Challenge, respBody.Session, nil
+}
+
+// queryAuthEndpoint queries the remote Habitat user's bff/auth endpoint
+func (c *client) queryAuthEndpoint(remoteHabitatUser *ExternalHabitatUser, sessionID string, proof string) (string, error) {
+	endpoint := c.constructRemoteHabitatEndpoint(remoteHabitatUser, "/habitat/api/node/bff/auth")
```
arushibandi, line 155: let's take this in the constructor so it doesn't get out of sync.

internal/resolvers/types.go
```diff
@@ -0,0 +1,11 @@
+package resolvers
+
+import "github.com/bluesky-social/indigo/atproto/crypto"
+
+type HabitatHostResolver func(did string) (habitatHost string, err error)
+
+type PublicKeyResolver func(did string) (publicKey crypto.PublicKey, err error)
```
arushibandi, line 7: thoughts ... I wonder if in the future we should do ... we could probably transform the `ExternalHabitatUser` into this
```
type Identity interface {
     HabitatHost() (habitatHost string, err error)
     PublicKey() (publicKey crypto.PublicKey, err error)
}
```
So that we can create a wrapper around and reuse this: https://github.com/bluesky-social/indigo/blob/bfc7f487e4ed5ece2c07a1bb309ba1d32dede33e/atproto/identity/identity.go#L116-L123

internal/bffauth/client.go
```diff
@@ -0,0 +1,217 @@
...
+	activeTokens map[string]string
+
+	// HTTP or HTTPS
+	scheme string
+
+	habitatHostResolver resolvers.HabitatHostResolver
+	publicKeyResolver   resolvers.PublicKeyResolver
+}
+
+func NewClient(clientDID string, privateKey crypto.PrivateKey) Client {
```
arushibandi, line 65: this needs to take in `habitatHostResolver` and `publicKeyResolver` no? otherwise how are the tests working ... ? if it ever runs it will panic

internal/resolvers/types.go
```diff
@@ -0,0 +1,11 @@
...
+
+type PrivateKeyResolver func(did string) (privateKey crypto.PrivateKey, err error)
```
arushibandi, line 9: why do we need this?

internal/resolvers/types.go
```diff
@@ -0,0 +1,11 @@
...
+
+// TODO resolve the DID from a host
```
arushibandi, line 11: doesn't atproto PDS do this already?

internal/bffauth/client.go
```diff
@@ -0,0 +1,217 @@
...
+	return &ExternalHabitatUser{
+		DID:       targetDID,
+		PublicKey: publicKey,
+		Host:      habitatHost,
+	}, nil
+}
+
+// getChallenge queries the remote Habitat user's bff/challenge endpoint
+func (c *client) getChallenge(remoteHabitatUser *ExternalHabitatUser) (string, string, error) {
+	endpoint := c.constructRemoteHabitatEndpoint(remoteHabitatUser, "/habitat/api/node/bff/challenge")
```
arushibandi, line 115: same here -- take in path via constructor as to not get out of sync

internal/bffauth/client.go
```diff
@@ -0,0 +1,217 @@
...
+	"github.com/bluesky-social/indigo/atproto/crypto"
+	"github.com/eagraf/habitat-new/internal/resolvers"
+	"github.com/pkg/errors"
+)
+
+type ExternalHabitatUser struct {
+	DID       string
+	PublicKey crypto.PublicKey
+	Host      string
+}
```
arushibandi, line 20: it's good practice to make types private if it's not used outside of the package

internal/bffauth/client.go
```diff
@@ -0,0 +1,217 @@
...
+		}, nil
+
+}
+
+type Client interface {
+	GetToken(targetDID string) (string, error)
+}
+
+type client struct {
+	clientDID string
```
arushibandi, line 52: nit: could just be `did` (so we don't have `client.clientDID` everywhere)

internal/bffauth/client.go
```diff
@@ -0,0 +1,217 @@
...
+
+	privateKey crypto.PrivateKey
+
+	activeTokens map[string]string
+
+	// HTTP or HTTPS
+	scheme string
```
arushibandi, line 59: hmmm why do we need this? can we just decide on one? (need https for security right)

internal/bffauth/client.go
```diff
@@ -0,0 +1,217 @@
...
+type Client interface {
+	GetToken(targetDID string) (string, error)
+}
+
+type client struct {
+	clientDID string
+
+	privateKey crypto.PrivateKey
+
+	activeTokens map[string]string
```
arushibandi, line 56: this is a nit but since everything is a `string` type, it would be nice if you can make
```
type did string
type token string
```
so it's clear what exactly is being stored, and easy to get a handle of what the code is doing even just from looking at types. some of these thing should be more public types so it's cool if you don't do it in this PR but going forwards I think this would be super helpful.

internal/bffauth/client_test.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	// Create client with test data
+	c := &client{
+		privateKey:          testKey,
+		activeTokens:        make(map[string]string),
+		scheme:              "http",
+		habitatHostResolver: FakeHabitatHostResolver(server.URL[7:]),
+		publicKeyResolver:   FakePublicKeyResolver(testPubKey),
+	}
```
arushibandi, line 87: using `NewClient` here would help catch for e.g. the bug where the resolvers are not passed in

internal/bffauth/client.go
eagraf: I ran into it because of the httptest server in the unit tests

internal/bffauth/client.go
eagraf: Yeah I feel like all of our PRs had this problem

internal/resolvers/types.go
eagraf: Leftover, removing

internal/resolvers/types.go
eagraf: Not really related to this PR, but it's something you need when doing OAuth, and I had a helper for it there. It would make sense to factor them out here (assuming OAuth was not an external package)

internal/bffauth/client.go
eagraf: Used the resolver to construct it, since it's constructed using the host

sashankg (changes requested):
We should have done some more research but I was going through the docs and realized that they already have a mechanism for this: https://docs.bsky.app/docs/advanced-guides/service-auth. It uses a jwt with a payload and then a signed version of the payload. The receiver just needs to verify with the public key from their did that the signature is correct. We can use https://docs.bsky.app/docs/api/com-atproto-server-get-service-auth to get the jwt. I think we just need to implement [verifyServiceJwt](https://github.com/bluesky-social/atproto/blob/f329c56454a178ef0d31671224f7370abe05142b/packages/xrpc-server/src/auth.ts#L72) in go (I couldn't find it but worth double checking)

sashankg:
implemented here: https://github.com/eagraf/habitat-new/pull/61

PR #58, opened 2025-04-06
controller cleanup

arushibandi:
- Delete the old `BaseNodeController` type in favor of `Controller2`.
- Also delete the `PDSClient` since we can either pass through directly to PDS from the front end or call via an `xrpc.Client`.
- Remove certificates from users -- we should just rely on DID. We don't use these anywhere.
- ^ As a result, we can remove the auth middleware. It's unused. We should rely on PDS did.
- Remove the `Enrich()` functions on transitions. Only one of these was being used and it could easily be moved to the Gen function for it. We should have stricter typing around the Gen functions and make all the transition types private but leaving this as a follow up.

Test Plan:
- I tested from the frontend making a new user and logging in my existing users and it works. 
- I also will fix up or add unit tests.


core/state/node/transitions.go
```diff
@@ -507,6 +474,18 @@ type AddReverseProxyRuleTransition struct {
 	Rule *ReverseProxyRule `json:"rule"`
 }
 
+func GenAddReverseProxyRuleTransition(t ReverseProxyRuleType, matcher string, target string, appID string) *AddReverseProxyRuleTransition {
```
arushibandi, line 477: In a follow up I would like to make all of these return `Transition` and the transitions themselves private types.

internal/node/controller/server.go
```diff
@@ -159,6 +160,90 @@ func (s *CtrlServer) GetNodeState(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	err = json.Unmarshal(slurp, &req)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
```
arushibandi, line 208: maybe time to make this part a helper function ... @sashankg how does glean make it look nice?

core/state/node/transitions.go
```diff
@@ -112,11 +104,17 @@ func (t *MigrationTransition) Validate(oldState hdb.SerializedState) error {
...
 
 type AddUserTransition struct {
-	Username    string `json:"username"`
-	Certificate string `json:"certificate"`
-	AtprotoDID  string `json:"atproto_did"`
+	User *User
+}
 
-	EnrichedData *AddUserTranstitionEnrichedData `json:"enriched_data"`
+func GenAddUserTransition(username string, did string) *AddUserTransition {
```
sashankg, line 110: uber nit: `Gen` is weird to me. Maybe `Create` or `New`

frontend/app/add-user/page.tsx
```diff
@@ -20,17 +20,21 @@ const Register: React.FC = () => {
...
+                    handle,
+                    password
+                }),
                 headers: {
                     'Content-Type': 'application/json',
-                },
-            });
+                }
+            })
+            console.log("resp", response)
```
sashankg, line 35: rm console.log

frontend/app/add-user/page.tsx
```diff
@@ -20,17 +20,21 @@ const Register: React.FC = () => {
         }
 
         try {
-            const response = await axios.post(`${window.location.origin}/habitat/api/node/users`, {
-                email,
-                handle,
-                password
-            }, {
+            // TODO: this should just pass through to the PDS xrpc endpoint -- unfortunately there's a bit more node setup that happens.
```
sashankg, line 23: yeah we should also use https://www.npmjs.com/package/@atproto/xrpc

frontend/components/authContext.tsx
```diff
@@ -54,7 +54,7 @@ export const AuthProvider: React.FC<{ children: ReactNode }> = ({ children }) =>
             }
 
             const fullHandle = parentDomain == "localhost" ? `${identifier}` : `${identifier}.${window.location.hostname}`;
-            const response = await axios.post(`${window.location.origin}/habitat/api/node/login`, {
+            const response = await axios.post(`${window.location.origin}/xrpc/com.atproto.server.createSession`, {
```
sashankg, line 57: can we switch this to `fetch` too

internal/node/controller/controller2.go
```diff
@@ -28,6 +32,7 @@ func NewController2(
 	pkgManagers map[node.DriverType]package_manager.PackageManager,
 	db hdb.Client,
 	proxyServer *reverse_proxy.ProxyServer,
+	pdsHost string,
```
sashankg, line 35: should we take in an xrpc client here instead? I'm thinking about if we want to start using the embedded pds methods we should use something more abstract

internal/node/controller/controller2.go
```diff
@@ -5,13 +5,16 @@ import (
 	"encoding/json"
 	"fmt"
 
+	"github.com/bluesky-social/indigo/api/atproto"
```
sashankg, line 8: is this the one where there's `atproto` and `agnostic` that have the same functions? if so what was the difference again and can we stick to one

internal/node/controller/server.go
sashankg: yeah we have wrappers for each error status that take a internal and user facing error

sashankg: imo leave as is for now until we have a better story for how to log and track errors

sashankg (approved):
this is awesome. I don't have the full context on some of these things so @eagraf should also review

frontend/app/add-user/page.tsx
arushibandi: leaving this as a follow up

internal/node/controller/controller2.go
arushibandi: Sadly it's because `atproto.RepoGetRecord` has a `util.LexTypeDecoder` and that's what I was having a tough time mocking out for tests

https://github.com/bluesky-social/indigo/blob/bfc7f487e4ed5ece2c07a1bb309ba1d32dede33e/api/atproto/repogetRecord.go

vs

https://github.com/bluesky-social/indigo/blob/bfc7f487e4ed5ece2c07a1bb309ba1d32dede33e/api/agnostic/repogetRecord.go

If there's a way to get around this though we should use `atproto`

internal/node/controller/controller2.go
arushibandi: oh hmmm. I was thinking we will have to construct multiple xrpc.Client's in here so we just want the pdsHost that is "tied" to this habitat node, but I guess it's fine if that's the xrpc.Client "tied" to the habitat node? unsure

PR #59, opened 2025-04-11
[FE] Tanstack router

sashankg:
Sorry for the framework thrash. Since we're not really using server rendering, Next.js makes things unnecessarily complicated. 

https://tanstack.com/router/latest seems well liked and handles everything client side. It also exposes a `loader` API which let's us handle async fetches in a nicer way than `useEffect`. It'll wait to render the page until the data is loaded. It should also make error handling easier. 

I also am adding https://picocss.com/ which should let us describe the page semantically and make it look good automatically. To some extent it limits our UIs but imo we shouldn't be wasting time with fancy UIs rn. I'm being extreme and not adding any css libraries to force us to stick to what pico provides. If we keep our components bare bones now we can more easily swap in a real design system later.

Some screenshots:
Screenshot 2025-04-11 at 6 54 08 PM
Screenshot 2025-04-11 at 6 55 49 PM
Screenshot 2025-04-11 at 6 55 02 PM
Screenshot 2025-04-11 at 6 54 39 PM


build/compose.yml
```diff
@@ -7,6 +7,7 @@ services:
       dockerfile: build/frontend.dev.Dockerfile
     volumes:
       - ${TOPDIR}/frontend:/app
+      - node_modules:/app/node_modules
```
sashankg, line 10: vite uses native code to bundle so we can't share the node_modules with local

frontend/src/api/node.tsx
sashankg, line 1: this is mostly unchanged

frontend/src/style.css
sashankg, line 1: made buttons green

arushibandi (approved):
yay love simplification

PR #69, opened 2025-04-28
simplify privi

arushibandi:
We realized with `atproto-proxy` header, we can remove a bunch of code and simplify paths. clients should set `atproto-proxy` header correctly by looking up where privi is hosted for a given DID.


arushibandi:
Duplicated by #73

PR #70, opened 2025-05-14
[OAuth] Initial buggy implementation

sashankg:
Inspired by https://github.com/eagraf/habitat-new/pull/45 but I didn't like the package's interface and wanted to separate out the OAuth logic from the server handlers. Also was just curious in understanding and thought it would be better for us to own the implementaiton anyway. Created an OauthClient interface that the api handlers wrap. 

Using https://github.com/gorilla/sessions to keep the implementation entirely stateless. All state information is stored on the user's browser as en encrypted cookie only we can read

## Known issues
- hard coded keys. 
  - client metadata key: this should be a persistent secret owned by the server. The OAuth provider can cache our client metadata and just randomly generating one at request time doesn't work sometimes.
  - sessions encryption key: this is less bad to generate at startup but means sessions won't persist across server restarts. This should also be persisted.
- Dpop Client is WIP. Haven't fully finished supporting actually making requests with an access token. we can do the e2e flow but the token isn't actually used yet
- Sometimes it just doesn't work. Still tracking down these issues, but we can probably move forward for now. Maybe the review can catch anything glaring I did


eagraf (commented):
Mostly unrelated to the PR, but I have a question about what happens when we have a PDS outside the habitat node. Does it still make sense for us to use the PDS for authenticating the user to interact with Habitat at that point?

internal/auth/login_handler.go
```diff
@@ -0,0 +1,231 @@
...
+		jwk, /*secretJwk*/
+	)
+	if err != nil {
+		log.Error().Err(err).Msg("error creating oauth client")
+		panic(err)
+	}
+
+	sessionStoreKey := make([]byte, 32)
+	rand.Read(sessionStoreKey)
+	sessionStore := sessions.NewCookieStore(sessionStoreKey)
```
eagraf, line 67: For testing, it'd help if the `sessionStore` was passed in through the constructor

internal/auth/login_handler.go
```diff
@@ -0,0 +1,231 @@
...
+		Key:       key,
+		KeyID:     "habitat",
+		Algorithm: string(jose.ES256),
+		Use:       "sig",
+	})
+	if err != nil {
+		log.Error().Err(err).Msg("error marshalling jwk")
+		panic(err)
+	}
+	oauthClient, err := NewOAuthClient(
```
eagraf, line 55: Same as the session store, would be nice to dependency inject the oauth client

internal/auth/oauth_client.go
```diff
@@ -0,0 +1,354 @@
...
+
+	return string(token), nil
+}
+
+type oauthProtectedResource struct {
+	AuthorizationServers []string `json:"authorization_servers"`
+}
+
+func fetchOAuthProtectedResource(i *identity.Identity) (*oauthProtectedResource, error) {
+	url, err := url.Parse(i.PDSEndpoint())
```
eagraf, line 189: Note that to my best knowledge this won't work with tailscale because the did document is not accessible on the tailnet domain name. My earlier workaround was to just manually pass in the PDS url since we assumed it would be on the Habitat node. This may not always be the case now that we're allowing external PDS

frontend/src/routeTree.gen.ts
```diff
@@ -25,6 +26,12 @@ const ServerRoute = ServerImport.update({
   getParentRoute: () => rootRoute,
 } as any)
 
+const OauthloginRoute = OauthloginImport.update({
```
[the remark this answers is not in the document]
sashankg: this is auto generated. the pathname is oauth_login

PR #74, opened 2025-05-17
[FE] Use @atproto/oauth-client-browser 

sashankg:
Add a separate auth path that uses the atproto browser oauth.


frontend/src/api/node.tsx
```diff
@@ -4,14 +4,8 @@ import * as node from '../../types/api';
 
 export const getNode = async (): Promise<node.GetNodeResponse> => {
   try {
-    const accessToken = Cookies.get('access_token');
-    if (!accessToken) {
-      throw new Error('No access token found');
-    }
-
```
arushibandi, line 11: this is just making it so you don't have to be logged in to see the node info right?

arushibandi (approved):

PR #80, opened 2025-06-22
Abandi/persist permissions state

arushibandi:
Privi now has a folder in `./habitat` and can persist permissions state across habitat process lifetimes.


internal/permissions/dummy_permission_store.go
```diff
@@ -9,21 +11,36 @@ type dummy struct {
 	permsByNSID map[syntax.NSID]xmaps.Set[syntax.DID]
 }
 
-func (d *dummy) HasPermission(didstr string, nsid string, rkey string, write bool) (bool, error) {
+func (d *dummy) HasPermission(didstr string, nsid string, rkey string, act Action) (bool, error) {
+	if act != Read {
```
sashankg, line 15: I think we should get rid of action entirely. I don't think we'll ever support writing to someone else's private data.

internal/permissions/permission_store.go
```diff
@@ -0,0 +1,143 @@
...
+		nsid string,
+	) error
+	RemoveLexiconReadPermission(
+		didstr string,
+		nsid string,
+	) error
+	ListReadPermissionsByLexicon() (map[string][]string, error)
+}
+
+type store struct {
```
sashankg, line 51: nit: storeImpl or even casbinStore

internal/permissions/permission_store.go
```diff
@@ -0,0 +1,143 @@
...
+func (p *store) HasPermission(
+	didstr string,
+	nsid string,
+	rkey string,
+	act Action,
+) (bool, error) {
+	return p.enforcer.Enforce(didstr, getObject(nsid, rkey), act.String())
+}
+
+// TODO: do some validation on input
```
sashankg, line 87: will we need to check for duplicates? Also is it possible to have conflicting permissions? don't need to figure out now but worth adding to the TODO

internal/permissions/permission_store.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+	return p.adapter.SavePolicy(p.enforcer.GetModel())
+}
+
+// TODO: do some validation on input
+func (p *store) RemoveLexiconReadPermission(
+	didstr string,
+	nsid string,
+) error {
+	// TODO: should we actually be adding a deny here instead of just removing allow?
```
sashankg, line 104: i think this will be based on the state of the existing permissions. For example, removing permission for a record key:
- If they have a read permission for the record, this should remove it
- If they have read permission for the namespace, this add a deny for that record

internal/privi/server.go
```diff
@@ -231,17 +255,99 @@ func (s *Server) getCaller(r *http.Request) (syntax.DID, error) {
...
+		http.Error(w, fmt.Errorf("no store found for caller %s", callerDID).Error(), http.StatusInternalServerError)
+		return
+	}
+
+	permissions, err := store.permissions.ListReadPermissionsByLexicon()
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	resp, err := json.Marshal(permissions)
```
sashankg, line 276: nit: you can do `json.NewEncoder(w).Encode(permissions)`

internal/privi/server.go
```diff
@@ -231,17 +255,99 @@ func (s *Server) getCaller(r *http.Request) (syntax.DID, error) {
...
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	if _, err := w.Write(resp); err != nil {
+		log.Err(err).Msgf("error sending response for ListPermissions request")
+	}
+}
+
+type editPermissionRequest struct {
```
sashankg, line 286: we need to setup the lexicon codegen

internal/permissions/permission_store.go
arushibandi: Added an issue for this

internal/permissions/permission_store.go
arushibandi: Ah okay so we need more granular permissions. I wonder if we should for now deny any requests on non-lexicons (i.e. record level permissions)

internal/permissions/dummy_permission_store.go
arushibandi: done!

internal/permissions/permission_store.go
arushibandi: done!

internal/permissions/permission_store.go
```diff
@@ -0,0 +1,143 @@
...
+				users.Add(perm[0])
+			}
+		}
+		res[strings.TrimSuffix(obj, ".*")] = slices.Collect(maps.Keys(users))
+	}
+
+	return res, nil
+}
+
+func getObject(nsid string, rkey string) string {
```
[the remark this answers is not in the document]
arushibandi: I removed this so that we only operate on nsid

internal/permissions/permission_store.go
```diff
@@ -0,0 +1,143 @@
...
+) (bool, error) {
+	return p.enforcer.Enforce(didstr, getObject(nsid, rkey), act.String())
+}
+
+// TODO: do some validation on input
+func (p *store) AddLexiconReadPermission(
+	didstr string,
+	nsid string,
+) error {
+	_, err := p.enforcer.AddPolicy(didstr, getObject(nsid, "*"), Read.String(), "allow")
```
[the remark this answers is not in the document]
arushibandi: removed

internal/permissions/permission_store.go
arushibandi: Added an issue

internal/permissions/permission_store.go
arushibandi: Added an issue for this

internal/permissions/permission_store.go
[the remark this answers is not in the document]
arushibandi: fixed

cmd/node/main.go
```diff
@@ -186,16 +190,45 @@ func main() {
...
-	sashankPerms := permissions.NewDummyStore()
-	sashankPerms.AddPermission("com.habitat.test", sashankDID)
-	perms[syntax.DID(sashankDID)] = sashankPerms
+	_, ok := perms[syntax.DID(sashankDID)]
+	if !ok {
+		perms[syntax.DID(sashankDID)] = permissions.NewDummyStore()
+	} else {
+		fmt.Println("found policies for sashank")
+	}
+	perms[syntax.DID(sashankDID)].AddLexiconReadPermission(sashankDID, "com.habitat.test")
```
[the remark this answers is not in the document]
arushibandi: why? i think we always want to do this even in the case it exists

sashankg (approved):

PR #96, opened 2025-07-19
pds: add to docker compose 

arushibandi:
Let's use the official pds and always run it in our docker compose since habitat is being packaged together with it anyways. A couple of things:
- This pds will be coupled to the permissions. So it is essentially a private data server.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,6 +66,10 @@ install:: $(DEV_HABITAT_PATH)/habitat.yml $(CERT_DIR)/dev_node_cert.pem $(CERT_D
 docker-build:
 	docker compose -f ./build/compose.yml build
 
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -66,6 +66,10 @@ install:: $(DEV_HABITAT_PATH)/habitat.yml $(CERT_DIR)/dev_node_cert.pem $(CERT_D
 docker-build:
 	docker compose -f ./build/compose.yml build
 
+# There's an issue when the front end has new dependencies that the package.json and pnpm-lock.yaml are not correctly updated. So rebuild with force no cache.
+docker-build-no-cache:
```
sashankg, line 70: shouldn't need this anymore

sashankg (commented):
is this still relevant if we directly use the repo implementation from indigo?

PR #100, opened 2025-07-21
Add some prompting advice for Cursor writing go tests

eagraf:
The surface area for testing the oauth implementation was very large. Cursor helped me write some basic unit tests following the rules commited here. Note that if we really want to test the oauth implementation well, we'd need somes sort of stronger integration or e2e suite.


eagraf:

- #100 ◀
    - #101
        - #102
            - #103
                - #104

PR #102, opened 2025-07-21
Reference correct PDS docker image arch

eagraf:
On ARM machines, the PDS image being pulled would not run properly. I've pushed an arm version of the PDS image to docker hub and adjusted the code to choose that if the host chip uses ARM.

PR #101, opened 2025-07-21
Fix npm install problems

eagraf:
Had some problems running the frontend.

PR #103, opened 2025-07-21
Allow for proxy rules to be configured at startup

eagraf:
Allow for custom proxy rules to be passed in from config files. This makes debugging and testing new apps really easy.For example, just run a node backend on your machine, and tell habitat what node and host its running on.


eagraf:

- #100
    - #101
        - #102
            - #103 ◀
                - #104

PR #104, opened 2025-07-21
XRPC Broker

eagraf:
[OAuth] Initial buggy implementation

Add proxy rules needed for PDS OAuth

Get PDS url from handle from internal PDS config if supplied

Fix some bugs with the DPoP client

- Reading of response body when use_dpop_nonce error occurs uses an unlimited buffer now
- Error handling for signing in the request
- Only read DPoP-Nonce header if a use dpop nonce error is detected
- DPoP private key session value is marshalled into a string (the key type wasn't allowed in the session)
- Auth token hash created using SHA256

Fix bugs in login_handler.go

- DPoP Gorilla session is saved without `defer` (was not being saved properly before)
- Fixed misnamed auth-session name
- Read access token from the token response properly

Hard coded configuration changes, fix later

Supply handle as login hint

Auth wip

Support custom PDS proxy rule

Inject sessionStore

Set a lot more info in the dpop session

htu claim is passed in

WIP PDS proxy rule

NewLoginHandler -> auth.GetRoutes

Fixup how htu is passed into Dpop client and overridden

Fixup dpopclient.Do

Add XRPC broker endpoint that forwards PDS requests to PDS with proper auth info

Remove PDS proxy rule

A JWKBuilder must be injected into the DPoP Client

Cleanup

Handle token refresh in test endpoint

Refactor jwks

Factor out DPoP session

Handle refresh token in broker

Fix some error strings in oauth client

More rb remove session store

Proper errror handling in session

Login handler test

Remove some hard coding

Oauth  client tests

Update login handler tests

Handler tests wip

Finish XRPC Broker tests

Routes test

Update logout to clear new cookies for gorilla sessions

Only allow Oauth login

Frontend fix wip

### DIFF
--- a/internal/auth/broker.go
+++ b/internal/auth/broker.go
@@ -29,7 +29,7 @@ func (h *xrpcBrokerHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// Note: this is effectively acting as a reverse proxy in front of the XRPC endpoint.
 	// Using the main Habitat reverse proxy isn't sufficient because of the additional
 	// roundtrips DPoP requires.
-	dpopSession, err := getExistingHabitatGorillaSession(r, w, h.sessionStore)
+	dpopSession, err := getCookieSession(r, w, h.sessionStore)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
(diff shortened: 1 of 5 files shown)


eagraf:

- #100
    - #101
        - #102
            - #103
                - #104 ◀

sashankg (approved):
so this will let apps reuse the habitat session to make changes to public pds. Fine with adding but might keep things simpler to just not support this for now. There probably are legit use cases for apps to interplay with public and private data but seems niche enough to not worry about rn

PR #113, opened 2025-08-03
[Privi] Multitenant permission store

sashankg:
We can simplify the permission store by supporting multiple users.

Update casbin model with `own(er)` key. Also removed action since we should only be supporting read.


internal/permissions/model.conf
```diff
@@ -1,10 +1,10 @@
 [request_definition]
- # sub(ject) = did, obj(ect) = nsid.rkey, act(ion) = read | write
-r = sub, obj, act
+ # sub(ject) = did, own(er) = did, obj(ect) = nsid.rkey
+r = sub, own, obj
 
 [policy_definition]
-# ef(fec)t = allow | deny
-p = sub, obj, act, eft 
+# sub(ject) = did or group, own(er) = did, obj(ect) = nsid.rkey, ef(fec)t = allow | deny
```
arushibandi, line 6: nit: can we remove 'group' until we support it

internal/permissions/model.conf
```diff
@@ -18,8 +18,7 @@ e = some(where (p.eft == allow)) \
 
 [matchers]
 # req subject must inherit from policy subject
+# req owner must match policy owner
```
arushibandi, line 21: i'm a little confused ... what's the difference between subject and owner?

internal/permissions/permission_store.go
```diff
@@ -100,39 +85,39 @@ func (p *casbinStore) AddLexiconReadPermission(
...
+	for _, policy := range policies {
+		lexicon := strings.TrimSuffix(policy[2], ".*")
+		// ignore denies for now
+		if policy[3] == "allow" {
+			res[lexicon] = append(res[lexicon], policy[0])
 		}
-		res[strings.TrimSuffix(obj, ".*")] = slices.Collect(maps.Keys(users))
 	}
 
+	log.Println(res)
```
arushibandi, line 120: nit: remove print

internal/privi/server.go
```diff
@@ -27,37 +27,27 @@ type PutRecordRequest struct {
...
 		stores: make(map[syntax.DID]*store),
 		dir:    identity.DefaultDirectory(),
 	}
-	for did, perms := range didToStores {
-		err := server.Register(did, perms)
-		if err != nil {
-			log.Err(err)
-		}
+	for _, did := range dids {
+		server.stores[did] = newStore(did, permissionStore)
```
arushibandi, line 46: hm ... this feels a bit weird but fine for now ... wonder if we want to be able to separate permission server from repo or if it should always go through

internal/permissions/permission_store_test.go
```diff
@@ -14,109 +13,123 @@ func TestBasicPolicy1(t *testing.T) {
...
-
-	for lex, perm := range perms {
-		require.Contains(t, exp, lex)
-		require.ElementsMatch(t, exp[lex], perm)
-	}
+	owner2Perms, err := ps.ListReadPermissionsByLexicon("did:owner2")
+	require.NoError(t, err)
+	require.Len(t, owner2Perms, 1)
 }
+
```
arushibandi, line 45: why did this test get removed?

internal/permissions/permission_store_test.go
```diff
@@ -14,109 +13,123 @@ func TestBasicPolicy1(t *testing.T) {
 	require.NoError(t, err)
 
 	// glob match on record key
-	ok, err := ps.HasPermission("did:1", "app.bsky.likes", "like1")
+	ok, err := ps.HasPermission("did:1", "did:owner", "app.bsky.likes", "like1")
```
arushibandi, line 16: can we simplify this example and just have `did:1` and `did:2` or `did:owner` and `did:grantee` or something? it's a little confusing rn and i think we lost some test cases

cmd/node/main.go
```diff
@@ -461,3 +418,33 @@ func initialState(
...
+	perms, err := permissions.NewStore(
+		fileadapter.NewAdapter(path.Join(policiesDirPath, "policies.csv")),
+		true,
+	)
+	if err != nil {
+		log.Fatal().Err(err).Msg("error creating permission store")
+	}
+
+	return privi.NewServer(
+		[]syntax.DID{},
```
arushibandi, line 447: shouldn't this take in a non empty list? otherwise we should get rid of this and just read off all the dids in the file? i.e. if you're served by this privi your did is in there somewhere?

internal/permissions/model.conf
sashankg: subject is the person or group the policy is affecting. Owner is the person that owns the record. For example in "Alice has permission for Bob's document", Alice is subject and Bob is owner

PR #115, opened 2025-08-09
Oauth client

eagraf:
Oauth client

Add oauth2 dep

Fix lint issues

Add some helpers

Fix names

Fix names

Remove filler tests

Fix test_utils

Fixes to auth server path


internal/node/api/router.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 type Route interface {
-	http.Handler
+	ServeHTTP(http.ResponseWriter, *http.Request)
```
sashankg, line 11: what's the difference?

internal/auth/oauth_client.go
```diff
@@ -0,0 +1,441 @@
...
+	RedirectUris            []string            `json:"redirect_uris"`
+	TokenEndpointAuthMethod string              `json:"token_endpoint_auth_method"`
+	TokenEndpointAuthSigner string              `json:"token_endpoint_auth_signing_alg"`
+	DpopBoundAccessTokens   bool                `json:"dpop_bound_access_tokens"`
+	Jwks                    *jose.JSONWebKeySet `json:"jwks"`
+}
+
+type OAuthClient interface {
+	ClientMetadata() *ClientMetadata
+	Authorize(dpopClient *DpopHttpClient, i *identity.Identity, handle *string) (string, *AuthorizeState, error)
```
sashankg, line 40: any reason `handle` needs to be a pointer?

sashankg (approved):

internal/auth/oauth_client.go
eagraf: It's misnamed, should really be `loginHint` which would be the same as a handle, but it should still function if we don't know the handle beforehand.

PR #120, opened 2025-09-14
privi persists to sqlite db

arushibandi:
Persist private data to a simple sqlite db with one table "records" and three columns:  [ did |  record key  | record ]

Trying to get us closer to a deployment of privi so we can start iterating quickly on top of that for actual apps.


cmd/node/main.go
```diff
@@ -339,11 +376,20 @@ func generatePDSAppConfig(
 				Matcher: "/xrpc",
 				Target:  "http://host.docker.internal:3000/xrpc",
 			},
+			/*
```
sashankg, line 379: let's add a proxy rule specifically for /.well-known/did.json so that we can restore this

cmd/node/main.go
```diff
@@ -217,55 +267,42 @@ func main() {
...
+		record TEXT,
+		PRIMARY KEY(did, rkey)
+	);`
+	_, err = priviDB.Exec(createTableSQL)
 	if err != nil {
-		log.Err(err).Msg("error on proxy-server shutdown")
+		log.Fatal().Err(err).Msg("unable to setup privi sqlite db")
 	}
-	log.Info().Msg("Gracefully shutdown Habitat proxy server")
+	defer priviDB.Close()
```
sashankg, line 298: doesn't this need to be in main?

internal/privi/repo.go
```diff
@@ -0,0 +1,153 @@
...
+			return err
+		}
+	}
+
+	bytes, err := json.Marshal(rec)
+	if err != nil {
+		return err
+	}
+	// Always put (even if something exists).
+	_, err = r.db.Exec("insert into records(did, rkey, record) values(?, ?, ?);", did, rkey, string(bytes))
```
sashankg, line 108: I think you can do `jsonb(record)` and sqlite stores a binary json representation

internal/privi/repo.go
```diff
@@ -0,0 +1,153 @@
...
+	return err
+}
+
+var (
+	ErrRecordNotFound       = fmt.Errorf("record not found")
+	ErrMultipleRecordsFound = fmt.Errorf("multiple records found for desired query")
+)
+
+func (r *sqliteRepo) getRecord(did string, rkey string) (record, error) {
+	rows, err := r.db.Query("select * from records where rkey = ? and did = ?", rkey, did)
```
sashankg, line 118: https://pkg.go.dev/database/sql#DB.QueryRow just returns 1 row

sashankg (approved):

PR #122, opened 2025-09-15
remove pds side car

arushibandi:
We don't use this and we've decided that in the short term, we will use public externally hosted PDS-es in combination with privi as personal data servers.


cmd/node/main.go
```diff
@@ -85,16 +81,14 @@ func main() {
 	}
 
 	// Generate the list of apps to have installed and started when the node first comes up
-	pdsApp, pdsAppProxyRules := generatePDSAppConfig(nodeConfig)
-	defaultApps, defaultProxyRules, err := nodeConfig.DefaultApps()
```
sashankg, line 89: let's keep defaultApps, right?

PR #124, opened 2025-10-04
privi standalone fixups

arushibandi:


cmd/privi/main.go
```diff
@@ -46,16 +74,29 @@ func main() {
...
+			fmt.Println("Got a request: ", string(x))
+			fn(w, r)
+		}
+	}
+
+	mux.HandleFunc("/xrpc/com.habitat.putRecord", logHandler(priviServer.PutRecord))
 	mux.HandleFunc(
 		"/xrpc/com.habitat.getRecord",
-		priviServer.PdsAuthMiddleware(priviServer.GetRecord),
+		logHandler(priviServer.PdsAuthMiddleware(priviServer.GetRecord)),
```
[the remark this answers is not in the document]
arushibandi: did this

cmd/privi/main.go
```diff
@@ -49,7 +79,22 @@ func main() {
 	}
 	priviServer := privi.NewServer(adapter, privi.NewSQLiteRepo(priviDB))
 
-	mux := http.NewServeMux()
+	mux := mux.NewRouter()
```
sashankg, line 82: i think no need to import `mux`. just do `Handler: loggingMiddleware(mux)`

arushibandi: oh nice

frontend/src/main.tsx
```diff
@@ -1,16 +1,15 @@
...
 import clientMetadata from "../client-metadata";
-
 // Import the generated route tree
 import { routeTree } from "./routeTree.gen";
-
 import reportWebVitals from "./reportWebVitals.ts";
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 import { BrowserOAuthClient } from "@atproto/oauth-client-browser";
 
+export const habitatProxy = "did:web:habitat.network#habitat"
```
sashankg, line 11: this should be in a separate file otherwise it's a circular dependency. not really an issue in typescript but better to avoid

sashankg (approved):

PR #125, opened 2025-10-04
[Privi] Decouple tailscale

sashankg:
- Added a script to funnel a port under arbitrary hostname using `tsnet`
- Added a [Procfile](https://devcenter.heroku.com/articles/procfile) (tldr; lets you enumerate processes that will run in parallel)
- added [foreman](https://github.com/ddollar/foreman?tab=readme-ov-file) (which lets you run a Procfile) build rule for privi dev

### DIFF
--- a/cmd/funnel/main.go
+++ b/cmd/funnel/main.go
@@ -1,0 +1,47 @@
+package main
+
+import (
+	"net/http"
+	"net/http/httputil"
+	"net/url"
+	"os"
+
+	"github.com/eagraf/habitat-new/internal/node/config"
+	"github.com/eagraf/habitat-new/internal/node/logging"
+	"tailscale.com/tsnet"
+)
+
+func main() {
+	// read command line args
+	port := os.Args[1]
+	hostName := os.Args[2]
+
+	logger := logging.NewLogger()
+
+	nodeConfig, err := config.NewNodeConfig()
+	if err != nil {
+		logger.Fatal().Err(err).Msg("error loading node config")
+	}
+
+	server := tsnet.Server{
+		Hostname: hostName,
+		AuthKey:  nodeConfig.TailscaleAuthkey(),
+	}
+	defer server.Close()
+
+	ln, err := server.ListenFunnel("tcp", ":443")
+	if err != nil {
+		logger.Fatal().Err(err).Msg("unable to listen")
+	}
+
+	localUrl, err := url.Parse("http://localhost:" + port)
+	if err != nil {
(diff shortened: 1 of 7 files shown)


cmd/funnel/main.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"github.com/eagraf/habitat-new/internal/node/config"
+	"github.com/eagraf/habitat-new/internal/node/logging"
+	"tailscale.com/tsnet"
+)
+
+func main() {
+	// read command line args
+	port := os.Args[1]
+	hostName := os.Args[2]
```
arushibandi, line 17: nit: probably should panic first is len(os.Args) < 3

cmd/funnel/main.go
```diff
@@ -0,0 +1,47 @@
...
+
+	logger := logging.NewLogger()
+
+	nodeConfig, err := config.NewNodeConfig()
+	if err != nil {
+		logger.Fatal().Err(err).Msg("error loading node config")
+	}
```
arushibandi, line 24: instead of using node config can we take in TailscaleAuthkey as a cli flag or argument?

i think node config is overloaded right now so trying to remove stuff from it. this should let us remove authkey

frontend/src/main.tsx
```diff
@@ -1,29 +1,28 @@
...
-      client_id: 'http://localhost?scope=atproto%20transition%3Ageneric',
-      redirect_uris: ['http://127.0.0.1:3001/'],
-      scope: 'atproto transition:generic',
-      token_endpoint_auth_method: 'none'
+  handleResolver: "https://bsky.social",
+  clientMetadata: __DOMAIN__
+    ? clientMetadata(__DOMAIN__)
+    : {
+      client_id: "http://localhost?scope=atproto%20transition%3Ageneric",
+      redirect_uris: ["http://127.0.0.1:3001/", "http://127.0.0.1:5173/"],
```
arushibandi, line 20: do we still need both of these?

arushibandi (approved):

cmd/funnel/main.go
sashankg: it'll panic anyway just won't have any err msg. If we end up adding any more code to this file I'll make it more user friendly

cmd/funnel/main.go
sashankg: TailscaleAuthkey is usually an env variable so I wanted to reuse that. Agree that config is overloaded and hard to use but will leave as in until we have a better solution

PR #127, opened 2025-10-04
privi server: error logging

arushibandi:
Log all errors for visibiliity (PDS does not forward error messages)

### DIFF
--- a/internal/utils/http.go
+++ b/internal/utils/http.go
@@ -10,7 +10,9 @@ var (
 	log = logging.NewLogger()
 )
 
-func LogAndHTTPError(w http.ResponseWriter, error string, code int) {
-	log.Error().Msg(error)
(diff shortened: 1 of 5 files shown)


internal/utils/http.go
```diff
@@ -0,0 +1,16 @@
...
+	"net/http"
+
+	"github.com/eagraf/habitat-new/internal/node/logging"
+)
+
+var (
+	log = logging.NewLogger()
+)
+
+func LogAndHTTPError(w http.ResponseWriter, error string, code int) {
```
sashankg, line 13: imo this should take the entire `err` as well as a `debugMsg string` so we know where the error is coming from. AI should do a good job auto completing the existing callsites

arushibandi: Ok i did this

PR #166, opened 2025-11-29
[clean up]: delete old node stuff + remove unused UI components + cleanup frontend auth/login

arushibandi:
Trying to prune as much as possible so we can switch over to the new repo in a clean state. Also trying to get the frontend into a place that's more flashy / explains what habitat does a bit.


cmd/funnel/main.go
```diff
@@ -29,14 +33,14 @@ func main() {
 	})
 	flag.Parse()
 
-	nodeConfig, err := config.NewNodeConfig()
-	if err != nil {
-		log.Fatal().Err(err).Msg("error loading node config")
+	tsAuthKey := utils.GetEnvString(tsAuthKeyVar, "")
```
sashankg: actually i think we can just remove this entirely. looks like tsnet already checks for this env var: https://pkg.go.dev/tailscale.com/tsnet#Server

frontend/src/auth.ts
```diff
@@ -6,14 +6,14 @@ const tokenLocalStorageKey = "token";
 const stateLocalStorageKey = "state";
 
 export class AuthManager {
-  handle: string | null;
+  handle: string | null = null;
```
sashankg: why do we need this? it gets set in the constructor, right?

sashankg (approved):

